PR #74, opened 2014-12-13
Display the code editor as preformatted text

cllns:
When going to the http://www.rust-lang.org/ with NoScript blocking all foreign assets, here's what my page looked like (since jsdelivr.net was blocked):
screen shot 2014-12-12 at 23 43 26

This PR is a one-line CSS fix (again without jsdelivr.net):
screen shot 2014-12-12 at 23 53 58


steveklabnik:
Thanks!

PR #76, opened 2014-12-18
Piping curl to sudo bash considered harmful

postmodern:
Following a discussion on twitter with @wycats about the safety, security and usability implications of curling a shell script into `sudo sh`, I made a few minor adjustments to the installation instructions.

1. Link to the [rustup.sh](https://static.rust-lang.org/rustup.sh) script so that users who wish to review the script can simply click on a hyper-link instead of manually copying the URL into their address bar.
2. Split the `curl | sudo sh` command into two separate commands. Now the `curl` command will download the `rustup.sh` script to a file. Again, this allows users to review the file before running it under sudo. This also leaves behind a "paper trail" for what and how Rust was installed onto their system.

I believe these changes respect the right of users to inspect how code is installed onto their systems, while being obtrusive to users who just want to install the latest Rust build right this second.


brson:
Closing because this is old. My current intent is to rewrite rustup.sh to prompt for sudo.

PR #87, opened 2015-01-14
Update default (source tarball) download URL

amaranth:
When you don't have Javascript or the script is unable to detect your platform the install button currently points to a link that does not exist.


alexcrichton:
Thanks!

PR #88, opened 2015-01-17
Change "Show me more!" to link to nightly intro

iKevinY:
The "Show me more!" link on the home page currently links to http://doc.rust-lang.org/master/intro.html, which 301 redirects to http://doc.rust-lang.org/nightly/intro.html; we can probably just link directly to the nightly intro.


alexcrichton:
Thanks!

PR #94, opened 2015-02-12
Fix warning about unknown or unused feature

ruud-v-a:
Currently the sample on rust-lang.org compiles with a warning.


huonw:
Thanks!

PR #98, opened 2015-02-20
Move <script> into <body>

frewsxcv:
\<script\> elements aren't allowed to live outside \<body\> or \<head\>. As a result, this will also make the front page valid HTML

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -78,7 +78,6 @@
     </ul>
   </header>
       {{ content }}
-  </body>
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

frewsxcv:
Woo! http://validator.w3.org/check?uri=http%3A%2F%2Fwww.rust-lang.org%2F&charset=%28detect+automatically%29&doctype=Inline&group=0

PR #102, opened 2015-03-23
Rename "Reference" → "Lang reference"

dhardy:
because I got fed up clicking "Reference" and not getting to the library reference I wanted!

Note: would have been "Language reference", but I couldn't figure out the CSS magic to make the columns wide enough!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,8 @@
+The Rust website ([www.rust-lang.org]())
+
+To preview your changes:
+
+    gem install jekyll
+    sudo yum install nodejs # install a javascript runtime; may not be necessary
+    PATH="$PATH:$HOME/bin" jekyll serve
+    # go to localhost:4000 or similar in a web browser!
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -0,0 +1,8 @@
+The Rust website ([www.rust-lang.org]())
+
+To preview your changes:
+
+    gem install jekyll
+    sudo yum install nodejs # install a javascript runtime; may not be necessary
```
huonw: Did you find that it was necessary?

dhardy: Yes, was for me.

Who cares if the README is a little sloppy, my guess is not many people are looking here, but this should get them started...

alexcrichton:
r? @steveklabnik 

I feel that as we haven't renamed the reference itself that "reference" is still the right name for it though...

steveklabnik:
Yeah, we re-named it to "Reference". I would prefer "Language Reference" over "Lang", as that's not even a word. But it feels redundant to me.

dhardy:
Seems to me that "Language Reference" would work fine if the columns were a little wider (how?). On second thoughts, just changing "API docs" to "Library API" would suit me.

brson:
@dhardy Perhaps we can do you latter suggestion of 'library api'.

This layout is going to have to change again soon once we add stable releases.

PR #106, opened 2015-04-09
Clean up smaller sized icons and add higher res icon to `favicon.ico`

Dr-Emann:
Added a 48x48 and 196x196 icon to `favicon.ico`. 196x196 is the
recommended size for google chrome on mobile for large shortcut icons.

Additionally, clean up the visibility of the rust icon at small sizes.


alexcrichton:
r? @brson 

Not sure if these came from a particular source or not and if we need to consider that.

brson:
Thanks!

Issue #107, opened 2015-04-11
Invalid value for version


whipsch:
The editor tries to make requests to the playpen using the version string 'master'.  The current playpen only accepts 'beta' and 'nightly'.

[Here](https://github.com/rust-lang/rust-www/blob/0443e47f0d1353a4b5d552a0475db0f356792264/_includes/editor.js#L79)

Not sure if we should be using beta or nightly.

brson:
Let's use beta.

PR #108, opened 2015-04-11
Improve playpen error handling

whipsch:
There was no case for handling an error returned by the playpen web application itself.  This commit ensures that future mishaps like #107 are more visible to users and are more likely to get reported due to the hard error message, instead of nothing happening.


brson:
Thanks!

PR #120, opened 2015-05-08
Updates for Rust 1.0

brson:
Do not merge yet.

This removes the asterisk about laundry, adds 1.0 links to the header, and a section of stable installers to install.html, adjusts the curl flags, removes .exe in favor of .msi.

One thing to note is that instead of adding additional stable links to the header I replaced beta with stable. It felt a bit overwhelming having a third set of duplicate links, and the beta docs seemed pretty dispensable.

cc @steveklabnik

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,7 +11,7 @@ title: The Rust Programming Language
           never segfaults,
           and guarantees thread safety.
           <br/>
-          <a href="http://doc.rust-lang.org/nightly/book/README.html">Show me more!</a>
+          <a href="http://doc.rust-lang.org/nightly/book/README.html">Show me!</a>
         </p>
       </div>
       <div class="col-md-4 install-box">
--- a/index.html
+++ b/index.html
@@ -8,8 +8,8 @@ title: The Rust Programming Language
         <p class="pitch">
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
-          prevents almost all crashes,
-          and eliminates data races.
+          never segfaults,
+          and guarantees thread safety.
           <br/>
           <a href="http://doc.rust-lang.org/nightly/book/README.html">Show me more!</a>
         </p>
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@ title: The Rust Programming Language
         <p class="pitch">
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
-          prevents almost all crashes<span class="asterisk">*</span>,
+          prevents almost all crashes,
           and eliminates data races.
           <br/>
           <a href="http://doc.rust-lang.org/nightly/book/README.html">Show me more!</a>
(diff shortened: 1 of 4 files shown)


brson:
@steveklabnik The link to explanation of unstable features is actually not that great. It links to the books 'nightly' page which is mostly another explanation of how to download nightly. What I really wanted was an explanation of release channels and how unstable features work.

brson:
This updates the pitch to 'Rust is a systems programming language that runs blazingly fast, never segfaults, and guarantees thread safety.'

cc @kmcallister

index.html
```diff
@@ -8,22 +8,22 @@
         <p class="pitch">
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
-          prevents almost all crashes<span class="asterisk">*</span>,
-          and eliminates data races.
+          never segfaults,
```
steveklabnik: I like changing this slogan, but 'never' is a bit strong here, given unsafe code. it's a hard caveat to fit in a slogan, though

install.html
```diff
@@ -5,43 +5,84 @@
 
     <div class="row install">
       <div class="col-md-offset-1 col-md-3 side-header">
-        <h2>1.0.0-beta.4</h2>
-        <h3>May 1, 2015</h3>
+        <h2>1.0.0</h2>
+        <h3>May 15, 2015</h3>
         <p>
-	  A feature-complete preview of the 1.0.0 release.
+	  The current stable release of Rust, updated every six weeks.
```
steveklabnik: What about something here to communicate that these updates are drop-in, easy upgrades?

brson:
Updated. @steveklabnik I tried to incorporate your suggestion as 'The current stable release of Rust, updated every six weeks and backwards-compatible.'

index.html
```diff
@@ -8,22 +8,22 @@
         <p class="pitch">
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
-          prevents almost all crashes<span class="asterisk">*</span>,
-          and eliminates data races.
+          prevents nearly all segfaults,
+          and guarantees thread safety.
```
steveklabnik: what about 'at compile time' here, since that's the magic?


sorry for suggesting even more coats of paint :(

install.html
```diff
@@ -5,43 +5,84 @@
 
     <div class="row install">
       <div class="col-md-offset-1 col-md-3 side-header">
-        <h2>1.0.0-beta.5</h2>
-        <h3>May 11, 2015</h3>
+        <h2>1.0.0</h2>
+        <h3>May 15, 2015</h3>
         <p>
-	  A feature-complete preview of the 1.0.0 release.
+	  The current stable release of Rust, updated every six weeks and backwards-compatible.
```
steveklabnik: :heart_eyes: :+1:

PR #123, opened 2015-05-15
Add security page

steveklabnik:
Not _quite_ ready to merge. We need the notifications mailing list, and for https://github.com/rust-lang/rust-www/pull/120 to link to it, so we'll probably merge this shortly after that lands.

### DIFF
--- a/security.html
+++ b/security.html
@@ -1,0 +1,170 @@
+---
+layout: default
+title: Security; The Rust Programming Language
+---
+
+<h2>Reporting a Bug</h2>
+
+<p>Safety is one of the core principles of Rust, and to that end, we would like to
+ensure that Rust has a secure implementation. Thank you for taking the time to
+responsibly disclose any issues you find.</p>
+
+<p>All security bugs in the Rust distribution should be reported by email to
+<a href="mailto:security@rust-lang.org">security@rust-lang.org</a>. This list
+is delivered to small security team. Your email will be acknowledged within 24
+hours, and you'll receive a more detailed response to your email within 48
+hours indicating the next steps in handling your report. If you would like, you
+can encrypt your report using <a href="rust-key.gpg.ascii">our public key</a>.
+This key is also <a
+href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xEFB9860AE7520DAC">On
+MIT's keyserver</a> and <a href="#key">reproduced below</a>.
+
+<p>This email address receives a large amount of spam, so be sure to use a
+descriptive subject line to avoid having your report be missed. After the
+initial reply to your report, the security team will endeavor to keep you
+informed of the progress being made towards a fix and full announcement. As
+recommended by <a href="https://en.wikipedia.org/wiki/RFPolicy">RFPolicy</a>,
+these updates will be sent at least every five days. In reality, this is more
+likely to be every 24-48 hours.</p>
+
+<p>If you have not received a reply to your email within 48 hours, or have not
+heard from the security team for the past five days, there are a few steps you
+can take:</p>
+
+<ul>
+        <li>Contact the current security coordinator (<a href="mailto:steve@steveklabnik.com">Steve Klabnik</a> (<a href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xDAE717EFE9424541">public key</a>)) directly.</li>
+        <li>Contact the back-up contact (<a href="mailto:andersrb@gmail.com">Brian Anderson</a> (<a href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0x16457A6368CFF26F">public key</a>)) directly.</li>
+        <li>Post on the <a href="https://internals.rust-lang.org/">internals forums</a> or ask in the #rust-internals IRC room on MozNet.</li>
+</ul>
+
+<p>Please note that the discussion forums and #rust-internals IRC channel are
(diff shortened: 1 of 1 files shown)


security.html
```diff
@@ -0,0 +1,170 @@
...
+responsibly disclose any issues you find.</p>
+
+<p>All security bugs in the Rust distribution should be reported by email to
+<a href="mailto:security@rust-lang.org">security@rust-lang.org</a>. This list
+is delivered to small security team. Your email will be acknowledged within 24
+hours, and you'll receive a more detailed response to your email within 48
+hours indicating the next steps in handling your report. If you would like, you
+can encrypt your report using <a href="rust-key.gpg.ascii">our public key</a>.
+This key is also <a
+href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xEFB9860AE7520DAC">On
```
jruderman: Please fix the unescaped & here and in the other pgp.mit.edu links

security.html
```diff
@@ -0,0 +1,170 @@
...
+these updates will be sent at least every five days. In reality, this is more
+likely to be every 24-48 hours.</p>
+
+<p>If you have not received a reply to your email within 48 hours, or have not
+heard from the security team for the past five days, there are a few steps you
+can take:</p>
+
+<ul>
+        <li>Contact the current security coordinator (<a href="mailto:steve@steveklabnik.com">Steve Klabnik</a> (<a href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xDAE717EFE9424541">public key</a>)) directly.</li>
+        <li>Contact the back-up contact (<a href="mailto:andersrb@gmail.com"></a> (<a href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0x16457A6368CFF26F">public key</a>)) directly.</li>
```
jruderman: This mailto link is missing anchor text.

security.html
```diff
@@ -0,0 +1,170 @@
...
+can encrypt your report using <a href="rust-key.gpg.ascii">our public key</a>.
+This key is also <a
+href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xEFB9860AE7520DAC">On
+MIT's keyserver</a> and <a href="#key">reproduced below</a>.
+
+<p>This email address receives a large amount of spam, so be sure to use a
+descriptive subject line to avoid having your report be missed. After the
+initial reply to your report, the security team will endeavor to keep you
+informed of the progress being made towards a fix and full announcement. As
+recommended by <a href="http://en.wikipedia.org/wiki/RFPolicy">RFPolicy</a>,
```
jruderman: Wikipedia link should use HTTPS

security.html
```diff
@@ -0,0 +1,170 @@
...
+these updates will be sent at least every five days. In reality, this is more
+likely to be every 24-48 hours.</p>
+
+<p>If you have not received a reply to your email within 48 hours, or have not
+heard from the security team for the past five days, there are a few steps you
+can take:</p>
+
+<ul>
+        <li>Contact the current security coordinator (<a href="mailto:steve@steveklabnik.com">Steve Klabnik</a> (<a href="https://pgp.mit.edu/pks/lookup?op=vindex&search=0xDAE717EFE9424541">public key</a>)) directly.</li>
```
jruderman: Having to combine a mailto link with a separate public key link seems suboptimal. Is there a way to include a public key in a mailto link? Or could you provide a web-based form?

bluss:
Can we spell out explicitly that breaking memory safety in rust code is a security issue? It's unclear to me if the intention is just rustc & other applications and regular security holes in them (what that would be) or something more directly relating to the safety properties of the language itself.

If I don't get it, others might be confused as well.

steveklabnik:
@bluss I wanted to make the topic of _what_ should be a security bug be an RFC we discuss, but have this page up in the meantime.

security.html
steveklabnik: There isn't a good way to include it, no.

steveklabnik:
@brson other than the few nits, are you okay with merging this?

brson:
@steveklabnik yes

security.html
steveklabnik: @jruderman why would this need to be escaped?

steveklabnik:
I merged this so we could see http://www.rust-lang.org/security.html , but it's not linked from anywhere yet. @jruderman , after I hear from you regarding the escaping issue, i'll make another PR linking it from the homepage.

security.html
reedloden: It needs to be "&amp;" instead of '&', as it's part of an HTML attribute. In fact, if you open up http://www.rust-lang.org/security.html in Firefox and view the source, you'll see the '&' is highlighted in red right now because it's invalid.

steveklabnik: I feel... really dumb. I guess I'm so used to frameworks generating this for me, that I always thought it would need to be escaped in text, but not as part of an `<a>`. https://validator.w3.org/check?uri=http%3A%2F%2Fwww.rust-lang.org%2Fsecurity.html&charset=%28detect+automatically%29&doctype=Inline&group=0 points this out too.

steveklabnik: continued in https://github.com/rust-lang/rust-www/pull/142

SimonSapin: For what it’s worth, it’s an authoring requirement in the HTML spec to escape `&` there, but there’s also a implementation requirement in the spec for parsers to fix it up so that the end the result is the same. So the concern is mostly theoretical.

PR #125, opened 2015-05-15
Changed reference to 'beta binaries' to 'stable binaries'

dguenther:
The install page refers to "beta binaries" under both the 1.0.0 section and the beta section. I changed the instance of "beta binaries" under the 1.0.0 section to be "stable binaries" instead.


alexcrichton:
Thanks!

PR #132, opened 2015-05-16
Fix play link not appearing in IE

### LINKED ISSUE #40: Icon doesn't render in IE11
Upon testing, the icon is valid in the page source, but not in the DOM Explorer. 

I think this is likely because of the `<i>` tags, which are rendered as italics or something in Internet Explorer.

From the Font-Awesome documentation:
> Font Awesome is designed to be used with inline elements (we like the `<i>` tag for brevity, but using a `<span>` is more semantically correct).

rust

dguenther:
Using innerHTML to remove elements in IE will wipe the content of the removed elements ([see here](http://www.quirksmode.org/dom/html/#t00)), which was causing the play link to not appear. This PR adds a method to remove the play link from the DOM before clearing the result div, allowing us to reuse it.

Tested in Safari 8.0.4, IE 11, Firefox 38, Chrome 42

Fixes #40

### DIFF
--- a/_includes/editor.js
+++ b/_includes/editor.js
@@ -61,6 +61,17 @@
   // Set initial size to match initial content
   updateEditorHeight();
 
+  // Safely remove all content from the result div
+  function clearResultDiv() {
+    // Clearing the result div will break our reference to
+    // the playlink icon, so let's save it if it exists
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #133, opened 2015-05-17
Use merged RfC link for governance RfC, not PR link

Manishearth:
The PR isn't rendered by default and shows a lot of discussion. While this discussion is important; in the context of the teams page people will more likely be interested in the workings of the teams and not the discussion up to it.

r? @alexcrichton


alexcrichton:
Good idea!

PR #138, opened 2015-05-19
Added a gemfile and installation instructions

aldeka:
So that other people aren't so foolish as to do `sudo apt-get install jekyll` and get the ancient broken version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 _site
+Gemfile.lock
+serve/
--- a/README.md
+++ b/README.md
@@ -1,0 +1,17 @@
+# rust-www
+
+This repository contains the source files for rust-lang.org.
+
+To install jekyll (and any other dependencies, if we get any):
+
+```
+    bundle install
+```
+
+To run this website as a developer:
+
+```
+    jekyll serve
+```
+
+This will serve the site at `localhost:4000`.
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1 +1,3 @@
-_site
\ No newline at end of file
+_site
+Gemfile.lock
```
steveklabnik: This actually should probably be in version control, we're an application, not a library.

README.md
```diff
@@ -0,0 +1,17 @@
+# rust-www
+
+This repository contains the source files for rust-lang.org.
+
+To install jekyll (and any other dependencies, if we get any):
+
+```
+    bundle install
+```
```
steveklabnik: If you do the ticks, you don't need an indent.

steveklabnik:
Thanks @aldeka ! Just one or two small nits, but you're right, this is _very_ helpful :)

.gitignore
aldeka: I'm not sure what you mean by this. It seems weird to me to have generated files in source control; then you get double the diffs (or worse, if you change a layout) every time you change something.

steveklabnik: Check out this post for the details: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

> When developing a gem, ... do not check your Gemfile.lock into version control... When developing an app, check in your Gemfile.lock,

PR #140, opened 2015-05-22
Removed indent from sides of install pages

aldeka:
None of the other pages have that offset, so, I removed it.

Also made the easy installation instructions a col-8 offset-4, so they line up with the download links.


steveklabnik:
:confetti_ball: :+1:

PR #159, opened 2015-08-04
Add timestamp of last site build to error page

edunham:
Useful for debugging purposes, and establishing whether Cloudfront
invalidation was done correctly. {{site.time}} inserts the current time when
the page is being built.

r? @alexcrichton


steveklabnik:
What about putting it into an HTML comment, so it's not directly displayed on-page?

edunham:
ok, it's a comment now. Also threw in a contributing guide for reasonably-obvious docs on the existance of the timestamp. 

Please don't merge until Travis is green:  I need to argue with it a bit about how not to run impossible commands on PRs, and I'll be rebuilding this commit as my test case.

brson:
Merge conflicts.

edunham:
Okay, I think in theory that we are good to merge now.

PR #173, opened 2015-08-13
Optimize the SVG version of the Rust logo

hannni:
Optimizes the SVG file from 7 033 Bytes to 3 542 Bytes (around 50%).
There are some changes that affect the geometry, but they all should be in the sub pixel range and not be visible.
A few more bytes could be saved by dropping the whitespace after the arc flags.
However probably people would complain because inkscape couldn’t read the file because of a bug (https://bugs.launchpad.net/inkscape/+bug/1284409).


alexcrichton:
GitHub at lease seems to think that they're pretty different?

[before](https://github.com/hannni/rust-www/blob/e00dc910e97785a8de794ead208173740d5e7f62/logos/rust-logo-blk.svg)

[after](https://github.com/hannni/rust-www/blob/68b2039b4d9ceda3348c04856037a36bf0ae2d63/logos/rust-logo-blk.svg)

Out of curiosity, what's the motivation for this? While saving 3K is great it seems a little small?

hannni:
Ah, I shouldn't have deleted the size attributes. Now they look the same for me both in Gecko (Firefox 40) and Webkit (Safari) both on OS X.

My motivation is that I'm currently writing a small rust program that optimizes the path strings of SVG. It's still pretty much work in progress, the xml part for instance is still missing.

alexcrichton:
Neat! In the interest though of preserving this as it was, I'm not sure how much of a benefit this is over what we have today?

brson:
I can't visually distinguish the difference, so I feel ok about this.

@hanni How was this compression done? Was it a standard tool.

hannni:
Page load time has a pretty big effect on users
(see e.g. https://blog.mozilla.org/metrics/2010/03/31/firefox-page-load-speed-part-i/).
This change probably won't have that big an effect
(the big culprits are ace.js and the fonts),
but I don't see the reason why not to save a few kilobytes.
The original file will still stay available through git.

Like I said before I'm currently developing a tool that does these compressions.
Parsing and outputting paths in an optimal way
relative/absolute commands, whitespace, dropping leading zeroes)
is as good as done and I used it for this file.
Combining of cubic splines still lacks an error estimation but I also used it for this file.
The rest of the optimizations were done by hand but I also want to support them automatically.
The code needs some cleanup and I barely worked on the xml part yet (only the path strings),
so I still haven't published the code.

brson:
Thanks @hanni!

Issue #182, opened 2015-08-28
RWIB Mission 8: Download Page Tweaks


brson:
[Link](https://internals.rust-lang.org/t/rusty-web-improvement-bureau/2572/1#download-page).

Rename from install.html to downloads.html, to match the 'Downloads'
link. The name 'Downloads' gives us room to expand to offer things
that are not installers.

To the beta channel, add an indication of which date it will become stable.

Add links to changelogs and/or release announcements, as well as the
docs for each channel.

Simplify labels by removing the words 'binaries' and 'installer'. This
is a meaningless distinction these days and just adds noise.

Two options for mac is too much choice w/o no obvious way to know
which one to pick. Remove one of them. I've started a poll to decide
which:
https://www.reddit.com/r/rust/comments/3i9dms/which_rust_installer_for_mac_do_you_use_poll/

Consider whether we need a better-organized archive to give access
bonus artifacts like the extra mac installer.

PR #187, opened 2015-09-11
Overhauled download page

brson:
This implements [RWIB 8](https://github.com/rust-lang/rust-www/issues/182).

[This is what it looks like](http://brson.github.io/rust-www/downloads.html).

The key things here are:

* The stable release number links to the release announcement
* There is only one Mac option
* There is a new 'docs' button for every channel
* The rustup.sh instructions are given more space so the blockquotes aren't cut off
* The beta channel indicates when it will be promoted
* Each channel indicates the associated version number

I'm reasonably happy with this. The only things aesthetically that I don't love are:

* The 'source' and 'doc' buttons are semantically different  from the other buttons, but visually not that distinct
* The three sets of rustup.sh instructions kinda feel 'heavy' and superfluous. I tried consolidating the beta/nightly instructions into a single space but didn't like the look of that either.


brson:
Another niggle: the spacing around the download labels is a little imperfect. I might extend the buttons to the left by another column.

brson:
cc @aturon @steveklabnik

brson:
@brson Consider making the 'source' and 'doc' buttons twice as wide and stacking them vertically. May make them more visually distinct.

alexcrichton:
:+1:, nice improvements!

The blue for the stable version header but black for the beta/nightly was a little off for me, perhaps the words "current stable release" could be a link to the release announcement? (or perhaps just the date?)

With only one download medium per platform we may also be able to drop the file extensions of what you're actually downloading as it's not giving you much other info now.

I like the idea of source/doc buttons being twice as wide, but perhaps the doc button should be a link in the description area? It's not necessarily a "let's download some stuff" button. (just a passing idea)

alexcrichton:
Oh and one other thought I just remembered, we may want to make sure to design in space to put links to MSVC builds for Windows, those are basically ready-to-go on nightly and basically-ready-to-go on beta right now, so we could plausibly be putting those links up pretty soon.

steveklabnik:
I like it.

aturon:
Nice work, @brson!

I agree with most of what @alexcrichton said, though I think we should keep the extensions in the labels. It would also be nice to move the labels closer to the download buttons, somehow. (Previously, the text was longer so the layout worked a bit better.)

Could the top release say "Stable" for consistency with the other channels? Also, I note that we don't give the patch number on the beta or nightlies, we should make this consistent as well.

I have mixed feelings about the centered `curl` command layout. Could you try having a wrapped line instead? I think this is one way that the current download page looks cleaner than the one proposed here.

brson:
I'm curious what people think of the name 'Downloads' and 'downloads.html' for this page instead of 'Install' / 'install.html'.

The thinking was that it made this page more flexible, but while I was reading this page I was thinking 'release channels ... this is a page about release channels'. Calling the page 'Release Channels' though is terrible.

brson:
Regarding the patch number my thinking was that there won't ever be a patch number for beta/nightly, and also that it makes those channels seem *less significant* than stable, like 'this isn't Rust 1.4.0; this is something 1.4-ish'. Similar reason for not putting 'Stable' in the first section - makes it stand out as different. I'll try some adjustments.

aturon:
@brson I prefer "Downloads" to "Install". Another option might be "Releases", but it's slightly less obvious that you go there to get downloads.

alexcrichton:
I don't have too much of a preference on naming, a slight inclination for "Install" over others, but not really that much.

llogiq:
Why not `get.html`? Apart from that, I know this is bootstrap CSS, but why cap the width at `1170px`? This would look squashed on 4k monitors, right? Even on one of my Full-HD monitors, it could make better use of screen estate.

steveklabnik:
> but why cap the width at 1170px? 

Really long line lengths are harder to read, so it's common to provide a maximum width.

llogiq:
Fine with me, but it either shouldn't be capped in pixels or adapt to higher-pixel-width screens. One solution would be to make the media query based on `em`s or `in`ches and define the used width in that unit.

steveklabnik:
I'm terrible at front-end, so I don't know what the _right_ thing is, but the studies on this are in terms of the length of words. So it would involve making sure that the right sized front at the right resolution isn't too long.

brson:
[Here's another revision](http://brson.github.io/rust-www/downloads.html). This is uploaded to the same place so you can't compare with the previous.

* Moved the blog link into the channel description.
* Changed the channel labels again to '1.3.0/beta/nightly'. I'm reluctant to put full three-point version numbers in the beta/nightly descriptions because those downloads only represent a release *series* - e.g. 1.4 beta is a beta of the series that will become 1.4.0.
* Removed the doc links since they aren't downloads.

I did try to compress the curl snippets to small width again by wrapping the lines. This isn't very usable though because it forces your copied commands to contain a line break, which sucks to edit after pasting. The command needs to fit on one line. We could keep the old design where the code block scrolls, but it's pretty unaesthetic. We could also make the command smaller by adding e.g. `rustup-nightly.sh` script.

alexcrichton:
Looks good to me!

brson:
I may try one more that adds the channel name and full version to all three.

PR #195, opened 2015-10-13
Add webpage for IDE work

nrc:
Deliberately not linked up for now. That'll come later, possibly after the website redesign.


brson:
Since it's not linked from anywhere else I just went ahead and merged it. Definitely needs to be integrated into the contributing section of the site at some point though.

PR #199, opened 2015-10-17
RWIB Mission 3 Contributing Page

efindlay:
This is an adaptation of [ncamerons](http://www.ncameron.org/rust.html) page.

Todo: 
- It should have a link to instructions on submitting bug reports.
- It should lead people to instructions for getting involved in RFC process.
- It should provide leads for getting involved in evangelism.
- It might highlight major initiatives, like IDE's. Such initiatives would need
stable landing pages to link to, like internals threads or GitHub issues.

PR #200, opened 2015-10-17
RWIB Mission 5 Documentation Page

efindlay:
This is a copy of the landing page in the documentation tree.

No changes to content, only cleanup of formatting.

I checked to see if there were any links in the documentation tree that required redirection in the documentation tree but there were none.

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -1,0 +1,147 @@
+---
+layout: default
+title:  Documentation
+---
+
+# Rust Documentation
+
+- [Getting Started][getting_started]
+- [The Standard Library][standard_library]
+- [Community &amp; Getting Help][community_and_help]
+- [Specification][specification]
+- [Tools][tools]
+- [FAQs][faqs]
+- [The Error index][error_index]
+- [Community Translations][translations]
+
+Welcome to the Rust documentation! You can use the section headings
+above to jump to any particular section.
+
+[getting_started]: #getting_started
+[standard_library]: #standard_library
+[community_and_help]: #community_and_help 
+[specification]: #specification
+[tools]: #tools 
+[faqs]: #faqs 
+[error_index]: #error_index 
+[translations]: #translations
+
+# Getting Started  <a name="getting_started"></a>
+
+If you haven't seen Rust at all yet, the first thing you should read
+is the introduction to [The Rust Programming Language][rustbook]. It'll
+give you a good idea of what Rust is like.
+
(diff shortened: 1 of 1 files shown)


brson:
I've pushed PRS #201 (community), #200 (docs), and #199 (contributing) to [my own gh pages so reviewers can see](http://brson.github.io/rust-www/). The first goal is to get the community page merged since it's the most developed.

The key pages under review are [community.html](http://brson.github.io/rust-www/community.html), [user_groups.html](http://brson.github.io/rust-www/user_groups.html), [documentation.html](http://brson.github.io/rust-www/documentation.html), and [contributions.html](http://brson.github.io/rust-www/contributions.html)

brson:
My focus is on the contributing page right now, but here are some thoughts on this one.

This is the documentation landing page for all of Rust; that includes the in-tree docs for all three channels, *possibly* docs for historical releases, the upcoming FAQ (hosted on the website), community and foreign translations, and crucially, this will also be the way we direct people to the vast amounts of non-official documentation for Rust, books.

Much of this page will subsume the [in-tree doc index](http://doc.rust-lang.org/index.html), but that page needs to continue to exist.

After this change I'd expect the in-tree index to contain the bare minimum: perhaps links to the books, the api docs and the error docs.

'Getting started' is definitely the right starting section for the doc page.

Next should be the 'Official documentation' or 'release documentation' (the in-tree docs). I'm not sure how to organize this offhand, whether we should relist all the in-tree doc components or just link to the in-tree landing page; we also have to link to multiple channels' documentation.

A lot of the unofficial docs we want people to find are available in [rust-learning](https://github.com/ctjhoa/rust-learning).

Ideas about categorization on this page:

* getting started
* getting help (maybe combined with the above) - this should not be as extensive as the existing, perhaps just #rust-beginners, stackoverflow and urlo.
* api - needs to link to various releases
* books and tutorials
* faq
* foreign language resources

Here's [Scala's page](http://scala-lang.org/documentation/) for reference - they interestingly begin with a menu of sorts for all the types of docs one might want, *then* go to the introductory stuff.

Some of the resources I think need to be exposed (all available from rust-learning):

* rustonomicon
* rust 101
* rust-learning
* entirely too many linked lists
* periodic table of rust types
* good practices for writing rust libraries

Sorry this has taken so long and is just a brain dump.

brson:
I've opened a new [PR](https://github.com/rust-lang/rust-www/pull/241).

PR #203, opened 2015-10-27
Add Windows -msvc links to the Downloads page.

briansmith:
It is very hard to find the native -msvc tools.

I didn't add links for stable 1.3 because my understanding is that 1.4 is the first version where -msvc will be recommended.


brson:
Thanks! I've made my own tweaks and included it in the [PR for tomorrow's release](https://github.com/rust-lang/rust-www/pull/204).

PR #209, opened 2015-10-29
Feature #rust-beginners prominently

brson:
We're trying an experiment with this new channel. Let's see how it goes if we promote it hard.

r? @steveklabnik


steveklabnik:
Sounds great!

PR #210, opened 2015-11-03
Retire Rachael from community team

brson:
r? @steveklabnik


steveklabnik:
Rachel is too busy for community team work with her job at IronWorks, and so asked us to retire until things change.

PR #212, opened 2015-11-04
Mention that VS2013 is needed

steveklabnik:
Fixes https://github.com/rust-lang/rust/issues/29479

r?

### DIFF
--- a/downloads.html
+++ b/downloads.html
@@ -184,7 +184,7 @@ title: Downloads &middot; The Rust Programming Language
 	</p>
 	<p>
 	  MSVC builds of Rust additionally require an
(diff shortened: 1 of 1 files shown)


downloads.html
```diff
@@ -184,7 +184,7 @@
 	</p>
 	<p>
 	  MSVC builds of Rust additionally require an
-	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio</a>
+	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio 2013</a>
```
petrochenkov: 2013 or newer (2015 is good too)

PR #213, opened 2015-11-05
Remove claim about "thread safety"

steveklabnik:
As https://github.com/rust-lang/rust/issues/26215 and others have pointed out, "thread safety" is a bad way to phrase this. When talking to people about Rust, the thing they've found most striking is compile-time errors, which is really unique to us, so it seems like a good proxy here.

r? @brson what do you think of this framing?

Fixes rust-lang/rust#26215

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,7 +9,8 @@ title: The Rust Programming Language
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
           prevents nearly all segfaults,
-          and guarantees thread safety.
+          and gives compile-time erorrs for
+          concurrency problems.
           <br/>
           <a href="http://doc.rust-lang.org/book/README.html">Show me!</a>
         </p>
--- a/index.html
+++ b/index.html
@@ -9,7 +9,7 @@ title: The Rust Programming Language
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
           prevents nearly all segfaults,
-          and gives compile-time erorrs for
+          and gives compile-time errors for
           concurrency problems.
           <br/>
           <a href="http://doc.rust-lang.org/book/README.html">Show me!</a>


index.html
```diff
@@ -9,7 +9,8 @@
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
           prevents nearly all segfaults,
-          and guarantees thread safety.
+          and gives compile-time erorrs for
```
oli-obk: erorrs -> errors

aturon:
Originally we had something about data races, which is what we actually provide a formal guarantee about (but is rather technical and not as compelling of a pitch).

I don't think the change here does much to clarify things, unfortunately.

We could say something like: "Enforces synchronization between threads" which is basically what data race freedom entails. That is, you are forced to do *some* synchronization, but aren't guaranteed that the synchronization does what you want.

brson:
I don't remember all permutations of this phrase but I think we settled on 'thread safety' as a succinct lay-explanation; people know or can intuit 'thread safety', not 'data races'. Also, as Niko points out in the thread, there is no commonly-accepted definition of thread safety.

I feel that 'and gives compile-time errors for concurrency problems' is too much of a mouthful for the one sentence master description of Rust.

* "and makes concurrency easy"
* "and runs on everything but the kitchen sink"

Not a lot of great ideas right now...

steveklabnik:
If we're all happy about this description, I'm happy to just throw my hands up and say "wontfix." I mean, it's marketing, not technical, so it can be close enough.

If that's the case, we should close https://github.com/rust-lang/rust/issues/26215

aturon:
@steveklabnik I feel pretty comfortable with that outcome. The current wording is defensible, and the list below it on the page gives more specific info wrt data races.

PR #216, opened 2015-11-06
Make note of checkbox on VS install

steveklabnik:
Someone come into #rust-beginners today, and they hadn't done this.

/cc @retep998

### DIFF
--- a/downloads.html
+++ b/downloads.html
@@ -185,7 +185,7 @@ title: Downloads &middot; The Rust Programming Language
 	<p>
 	  MSVC builds of Rust additionally require an
 	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio 2013 (or later)</a>
-	  so rustc can use its linker.
+	  so rustc can use its linker. Make sure to check the "C++ tools" option.
 	  No additional software installation is necessary for basic use of the GNU build.
 	</p>
 	<p>


steveklabnik:
http://i.imgur.com/aW5Ai3d.png is a screenshot, maybe we should use it?

steveklabnik:
http://i.imgur.com/6kSmaUe.png

alexcrichton:
I tend to be a little wary of snippets like this, it's 100% guaranteed to go out of date as VS changes their installation procedures and various pieces here and there, and this is quite a prominent location to put this sort of info. This seems like it may be better as an error message in the compiler or something like that.

retep998:
It's a good idea to point out wherever we tell people to install VS that they should make sure they also install the C++ components. Waiting until they've installed VS and installed Rust and then restarted their computer and finally tries to build something only to get an error spit at them that they installed VS wrong isn't very nice.

steveklabnik:
I share the concern about bitrot, but there is a tension here, as the download is huuuuuge

brson:
This is ultimately the kind of thing we want rustup/cargo to detect and make suggestions about. Every additional word we have to add to explain wtf to do to get Rust running also makes me super-duper-sad.

I do think though that if this is where we tell people to get VS, then this is where we have to make it clear exactly what do do.

In a future world where rustup.exe does everything, then rustup.exe can explain it in a platform-dependent way.

As to the picture, I simply will die if we plaster that in the middle of this page. Maybe an inline hyperlink to a lightbox, but still - grrrrr.

steveklabnik:
I wasn't suggesting embedding, just a link or lightbox or something, yeah

brson:
@steveklabnik how about merging this as is and figuring out the pic in another?

steveklabnik:
Sure thing :)

brson:
cc @Diggsey I mentioned above how rustup.exe/multirust can save us from lots of exposition about windows by detecting whether VS exists properly.

PR #217, opened 2015-11-06
Rename "Community" nav header to "Resources"

brson:
This way "Community" isn't repeated.

r? @steveklabnik

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -46,9 +46,6 @@
       </li>
       <li class="col-xs-4 col-md-2"><h2>Resources</h2>
         <ul>
-          <li>
-            <a href="http://this-week-in-rust.org/">News</a>
-          </li>
 	  <li>
 	    <a href="community.html">Community</a>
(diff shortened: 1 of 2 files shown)


brson:
cc @aturon

brson:
This is temporary until the docs page is redesigned, but it will be a long temporary.

aturon:
SGTM. The repeated "Community" is a bit weird at the moment. "Resources" is generic but at least avoids the repetition.

One question: why is `News` called out currently, given that it's also linked in Community? With this refactor, I'd at least suggest moving the `Community` link to the top, since it contains the largest set of leads.

brson:
@aturon The question about 'News' was also brought up on IRC. It's fairly arbitrary.

I've removed it and instead added the security policy.

brson:
cc @steveklabnik Adding security policy to front page.

steveklabnik:
Really glad to see this get more attention :+1:

PR #225, opened 2015-11-23
State that Rust prevents segmentation faults

### LINKED ISSUE #191: Rust's short description on the home page is a bit weird.
>Rust is a systems programming language that runs blazingly fast, prevents nearly all segfaults, and guarantees thread safety. 

I think this should say that it prevents all segfaults – otherwise it's inconsistent with the last part, that it guarantees thread safety. You can break both invariants by using `unsafe`. The other possibility would be to phrase it like "prevents nearly all segfaults and almost guarantees thread safety".

Also, shouldn't it say data-race freedom instead of thread safety?

tbu-:
The rest of the home page talks about safe Rust:
- guarantees thread safety
- guaranteed memory safety
- threads without data races

Fixes #191.

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@ title: The Rust Programming Language
         <p class="pitch">
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
-          prevents nearly all segfaults,
+          prevents segfaults,
           and guarantees thread safety.
           <br/>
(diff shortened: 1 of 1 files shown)


steveklabnik:
Yeah, this is a good point, if we're assuming safe rust everywhere else, might as well do it here too.

PR #228, opened 2015-11-24
Remove a superfluous space

erickt:
Any reviewers should be extra safe and confirm I'm not changing @steveklabnik's pgp key link :)


steveklabnik:
I'll do it myself :)

PR #230, opened 2015-12-06
Add #rust-networking to the IRC channel list

mrmonday:
I don't know if this is appropriate, since we're not officially endorsed, but if you're happy to have us we'd love to be on the list.


alexcrichton:
Thanks!

PR #231, opened 2015-12-08
Update meta tag description to match index page

56quarters:
Noticed that the meta tag description no longer matches the blurb on the homepage so I've changed it. Thanks!


alexcrichton:
Thanks!

PR #239, opened 2015-12-17
Add missing copy of Security Team's public key

felixc:
The Rust Security Policy page (https://www.rust-lang.org/security.html)
links to the team's public key in three ways: a copy embedded at the
bottom of the page, a link to the MIT keyserver, and a link to a hosted
copy (https://www.rust-lang.org/rust-security-team-key.gpg.ascii). This
last link is currently broken, because the referenced file is not
present in the repository.

The version added in this commit is copied from the embedded one.


alexcrichton:
Thanks!

Issue #179, opened 2015-08-28
RWIB Mission 5: Documentation Page


brson:
[Link](https://internals.rust-lang.org/t/rusty-web-improvement-bureau/2572/1#documentation-page).

Today, [the documentation landing
page](http://doc.rust-lang.org/nightly/) lives in the Rust tree, and
is generated for each release channel. As Rust grows though, more
documentation lives out of tree and needs to be discoverable.

Let's reorganize this so that the website itself is the main landing
page for finding docs.

Start by moving the contents of the existing documentation index into
rust-www, reducing the existing doc index to links to the in-tree
docs, as well as a link to the website for more information.

Add a link to the new website-based [FAQ](#faq).

Brainstorm about other documentation we might want to link to and what
standards of quality and stability we might require before officially
endorsing them.

efindlay:
I would like to take this on after #193.

brson:
@efindlay You got it! Thanks!

efindlay:
It looks like my first step should just be a pure refactor with no visible changes (except for URL). I'm guessing the tricky part will be finding and changing links to the current documentation landing page.

efindlay:
Set of possible links:

https://doc.rust-lang.org
https://doc.rust-lang.org/
https://doc.rust-lang.org/index.html
https://doc.rust-lang.org/nightly
https://doc.rust-lang.org/nightly/
https://doc.rust-lang.org/nightly/index.html
https://doc.rust-lang.org/stable
https://doc.rust-lang.org/stable/
https://doc.rust-lang.org/stable/index.html

efindlay:
Grepped with,

git grep -En 'doc.rust-lang.org[^/]'
git grep -En 'doc.rust-lang.org/[^a-z]'
git grep -En 'doc.rust-lang.org/index.html'
git grep -En 'doc.rust-lang.org/nightly[^/]'
git grep -En 'doc.rust-lang.org/nightly/[^a-z]'
git grep -En 'doc.rust-lang.org/nightly/index.html'
git grep -En 'doc.rust-lang.org/stable[^/]'
git grep -En 'doc.rust-lang.org/stable/[^a-z]'
git grep -En 'doc.rust-lang.org/stable/index.html'

efindlay:
Oh. Only links were part of #![doc ... html_root_url = https://doc.rust-lang.org/nightly/ which doesn't need to change, does it?

brson:
This is done.

PR #229, opened 2015-11-30
Contributions page

brson:
Based off of [@efindlay's original PR](https://github.com/rust-lang/rust-www/pull/199), which was based off [ncameron's work](http://ncameron.org/rust.html).

[Rendered](http://brson.github.io/rust-www/contribute.html).

The idea here is to quickly provide the most important links (e.g. the second paragraph reroutes people to bug reporting), subdivide the deluge of information in a way that is digestible (six sections - issue triage, docs, community building, tooling, libs, compiler), then guide people to *other sources of information*. I've used prose in an attempt to provide context for newbies to understand the larger picture. Sometimes the same information is provided in different sections, from different perspectives - e.g. fixing easy issues vs. writing easy issues.

Ideally, there is an easy path from here to all relevant information for contributors, but that is not quite true yet, and there are types of information that are just not available. What I *want* this page to do is to link to other, 'living', landing pages for teams, initiatives, topics, etc., and I've tried to do that as much as possible, but there are some categories of information that are not organized anywhere, so sometimes there is more exposition here than I'd like.

This page occasionally links to, and thus endorses, unofficial projects of interest. I want to be clear though that it is not a free-for-all and we are not open to listing arbitrary projects here.

With the quantity and specificity of info here it'll be prone to bitrot and need periodic attendance.

Problems:

* The prominent link to 'an overview of the organization, processes, and policies' links somewhere uncomprehensive. I'd like to write this, but there's a lot of info to cover and I want a break from writing for a bit.
* Topics of blogging and translation aren't covered.
* #rust-libs may not be the right place to direct people who are interested in out-of-tree libs.

I've run out of steam on this and want to get other eyes on it. The libs and tooling sections were the last I hit and it shows.

Although they are very different, my main influence was [Scala's page](http://scala-lang.org/contribute/).

Fixes https://github.com/rust-lang/rust-www/issues/177.

cc @aturon @steveklabnik

PR #202, opened 2015-10-22
Added Frequently Asked Questions Page

AndrewBrinker:
This commit is the first draft of a new Frequently Asked Questions (FAQ) page for the Rust website. It provides answers for all but 4 of the 130 questions asked by the community, which can be found on the tracking issue #181.

The answers could use some review, as could the design and organization. I've tried to provide clear, concise answers that more often than not point people to the relevant portion of the Rust book.

Thoughts? Changes?

Issue #175, opened 2015-08-28
RWIB Mission 1: 


brson:
[Link](https://internals.rust-lang.org/t/rusty-web-improvement-bureau/2572/1#front-page)

The amount of content we need to present is outgrowing our nav
bar. Community links are growing and there are lots more links we need
to expose but no room.

Completely restructure the nav bar as a few 'category' links that lead
to their own pages:

* Documentation
* Community
* Downloads
* Contributing

The content of these pages is discussed in further missions
below. Because this adds 'downloads' to the navbar, we also need to
remove the redundant 'Other Downloads' button.

There are some other important links that need to be integrated
into the front page:

* crates.io - It takes a *lot* of clicks to discover that crates.io
  exists today! I don't know the best way to integrate this, but one
  possibility is to call the crate ecosystem out in the feature list
  and hyperlink there.
* play.rust-lang.org - To discover play.rlo you have to click 'Run',
  then the 'fold-out' icon. This is a useful enough newbie tool
  that it ought be completely obvious how to find it. Again, no great
  ideas about where this link lives in the current design.
* stability guarantee - I like the idea of adjusting the sales pitch /
  feature list to tout stability and making this an inline link there.
  This isn't written yet, discussed below in another missions.
* security policy - The [page
  exists](http://www.rust-lang.org/security.html) but isn't linked.
* GitHub - via banner is fine.

If these can't all be integrated organically into the flow of the page
then you might consider adding them to a new footer.

Consider making the feature list link to various blog posts. Make sure
the page style continues to attract attention to the more crucial
links we already have.

Propose tweaks to the design to accomplish all these goals.

nagisa:
I personally like the current simplistic design of the frontpage and believe that we should try to preserve that.

> Completely restructure the nav bar as a few 'category' links that lead to their own pages:

Sounds a lot like what happened to haskell.org.

> Because this adds 'downloads' to the navbar, we also need to remove the redundant 'Other Downloads' button.

Rather than that I think the navigation bar looking something like

```
[logo] [Documentation] [Community] [Contributing] [Download for OS | Other downloads]
```

is better.

Usually when menus become unmanageable, some sort of sub-menus are employed, such as used by the [Python] website. We could do something similar as well. I like the way Python’s website splits the organisational links (e.g. PyPI and PSF) and links related to implementation of Python itself into two separate menus. While we probably don’t have enough links to consider that yet, it might point to the correct approach for us.

[Python]: https://www.python.org/

[...]

PR #241, opened 2015-12-23
Add a documentation page. Restructure navigation

brson:
This adds a documentation page to the website, to replace the in-tree landing pages, and restructures the top level navigation to have only 'Documentation', 'Community', 'Downloads', and 'Contribute' links.

The security, legal and FAQ pages are all linked from the docs page.

To deal with nightly and beta docs I just included a section called 'nightly and beta documentation' that goes only to the in-tree landing page for those, the thinking being that they are less important, and for more knowledgeable users. So I did not try to duplicate all the relevant information for all three channels ala the downloads page.

For non-english resources I deferred to [rust-learning](https://github.com/ctjhoa/rust-learning) temporarily, though I'd prefer to use the [forge](https://internals.rust-lang.org/t/dungeon-rust-lang-org-secondary-developer-docs/2896) for that. I [opened a PR](https://github.com/ctjhoa/rust-learning/pull/26) against rust-learning to include all the book translations.

For other useful resources I again deferred to rust-learning. To keep things non-overwhelming I like having secondary resources on another page, though perhaps rust-learning will not be that page in the future.

You can see this deployed, along with the contributions page [here](http://brson.github.io/rust-www/index.html).

Needs to wait on the [contributions](https://github.com/rust-lang/rust-www/pull/229) page to merge, and if it merges before the [FAQ](https://github.com/rust-lang/rust-www/pull/202) then it needs to be edited to make the FAQ link work.

Fixes https://github.com/rust-lang/rust-www/issues/179. cc https://github.com/rust-lang/rust-www/issues/175

cc @efindlay @aturon @steveklabnik


brson:
I used [Go](https://golang.org/doc/) and [Scala](http://scala-lang.org/documentation/) as references.

aturon:
The docs page looks pretty good to me. I just have a couple of overall worries:

- The various policy docs now feel a bit buried. Granted, these things are often given as small links at the bottom of a page anyway, so it's not clear that this is much worse -- my main concern is I wouldn't think of "Documentation" as the place to find them. At the very least, I think we should have a distinct "Policy" subsection of documentation, and ultimately we'll be able to add docs on e.g. the governance setup and RFC process there.

- The layout of the front page seems... off... somehow. In particular, the whitespace around the top links seems excessively, and none of the elements in the main body align with each other. I'm struggling to be more constructive -- maybe I should just fiddle with it on a branch and see if I can demonstrate what I have in mind.

brson:
I've pushed a commit to split out a 'project policies' section. Adding a section on 'policies' generally suggests we need more links there. One possibly is the link explaining Rust governance and processes that I have mentioned to you previously.

I've been thinking about adding a footer across the site, particularly to explain how to submit changes to the website itself, possibly to explain the copyright of the website itself. We could add other legal links there, but frankly I think we should not, because they are not so important that that they should be available from every page of the website.

I'm going to be doing some work on integrating the FAQ styling with the rest of the site. Part of this may be touching how the header is separated from the content. I'll see if anything there suggests ways to make this header look more balanced.

brson:
I've pushed a change to my deployment that adjusts the style of the site to accomodate the styling of the faq. The changes are described [here](https://github.com/rust-lang/rust-www/pull/202#issuecomment-168873278). Part of this adds a border element to most pages near the navbar, and it also reduces the whitespace around the navbar on the index page.

aturon:
@brson Looking very good! I think the separate policy section works well, and agree that it's a natural place to talk about governance, etc. (Though I don't think we have to do that for this initial round of website updates).

My only (very small) nit is that the horizontal rule separating the header from the rest seems thinner than the rule separating the different sections, which felt a bit jarring.

But overall the rules and large section headers feel pretty good as a way to present a lot of information in a manageable and simple way.

brson:
On reddit /r/killercup suggested that the [syntax index](https://github.com/rust-lang/rust/pull/30814) should be advertised more widely. I already added it to rust-learning, and I'm tempted to add it to the references section.

aturon:
@brson Sounds like a good idea to me.

brson:
@aturon My latest version uses a consistent-width horizontal rule everywhere.

This PR is subsumed by [that].

PR #256, opened 2016-01-15
Fix small typo in faq.md

joliv:
Under "Other Languages" > "How do Rust traits compare to Haskell typeclasses?" foo and bar was mistyped as foo and boo :)


alexcrichton:
Thanks!

Issue #246, opened 2016-01-14
Enable syntax highlighting for faq


brson:
The new FAQ has a bunch of Rust code that isn't highlighted.

azerupi:
Are you guys already using a library for that?
I use [highlight.js](https://highlightjs.org/) for mdBook and it works well.

steveklabnik:
It's rendered server-side, Jekyll does it.

azerupi:
Is it just a matter of replacing triple back ticks with the tags below?

```
{% highlight rust %}

{% endhighlight %}
```

azerupi:
I have investigated the html output and code blocks are rendered with the correct html for styling:

```html
<pre><code class="language-rust" data-lang="rust"><span class="k">fn</span> <span class="n">is_odd</span><span class="p">(</span><span class="n">x</span><span class="o">:</span> <span class="kt">i64</span><span class="p">)</span> <span class="o">-&gt;</span> <span class="kt">bool</span> <span class="p">{</span>
    <span class="k">if</span> <span class="n">x</span> <span class="o">%</span> <span class="mi">2</span> <span class="o">!=</span> <span class="mi">0</span> <span class="p">{</span> <span class="k">true</span> <span class="p">}</span> <span class="k">else</span> <span class="p">{</span> <span class="kc">false</span> <span class="p">}</span>
<span class="p">}</span>
</code></pre>
```

So it's highly probable that the only problem is a missing stylesheet.

brson:
@azerupi Great! Will you submit a PR?

azerupi:
Well I can't test it locally, so I can't really pinpoint the specific problem. But I know for sure that it's a missing css file. I will try to take a look, but if someone else can beat me to it, please do :)

PR #258, opened 2016-01-15
FAQ - Replaced triple backticks with jekyll highlight tags

azerupi:
:warning: I am not sure this works! I can't test it locally because I always have a dependency error:

```
Dependency Error: Yikes! It looks like you don't have redcarpet or one of its dependencies installed. In order to use Jekyll as currently configured, you'll need to install this gem. The full error message from Ruby is: 'cannot load such file -- redcarpet' If you run into trouble, you can find helpful resources at http://jekyllrb.com/help/! 
  Conversion error: Jekyll::Converters::Markdown encountered an error while converting 'community.md':
                    redcarpet
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    redcarpet
```

I tried to install `redcarpet` and `rouge` like advised on the internet but that didn't help. Can someone test this and see if it works?

It is supposed to enable code highlighting in the FAQ (#246)

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -268,19 +268,19 @@ When should I use an implicit return?
 
 Rust is a very expression-oriented language, and "implicit returns" are part of that design. Constructs like `if`s, `match`es, and normal blocks are all expressions in Rust. For example, the following code checks if an [`i64`][i64] is odd, returning the result by simply yielding it as a value:
 
-```rust
+{% highlight rust %}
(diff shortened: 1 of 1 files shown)


brson:
I tried this - highlighting with redcarpet and rouge - and it did not generate any colors. I will try to investigate again soon.

azerupi:
Ah ok :)
Good luck then!

PR #260, opened 2016-01-15
Fix relative links on error page

ollie27:
This should fix the broken images and links on urls like:
https://www.rust-lang.org/fail/


alexcrichton:
cc @brson, does this make developing locally any harder? or is that still a pain regardless?

brson:
This should be fine.

PR #269, opened 2016-01-15
Correct StackOverflow -> Stack Overflow in FAQ

jeremybanks:
Per the "Proper Use of the Stack Exchange Name" section of http://stackexchange.com/legal/trademark-guidance:

> As a name, Stack Overflow, is always written "Stack Overflow" (two words, capital letters).


steveklabnik:
Thank you!!!

PR #272, opened 2016-01-15
Removed bold from links to improve readability

listochkin:
When Brian did [the documentation page overhaul](https://github.com/rust-lang/rust-www/commit/893021c95c286e637add50f7a952ecbbb893a364) he made most links in it bold. However, on OS X these small bold letters look fuzzy and are harder to read. The problem is more apparent in Firefox that tends to render glyph strokes wider than other browsers. Here I've put up the page fragment showing bold and regular fonts being used for links as it's rendered on my machine:

Firefox render for bold and regular links

Also, other pages on Rust website don't use bold for links within text body, and as it is now the documentation page looks out of place.

I know that the whole thing is highly subjective, so if Brian or anyone else feels strongly about the new style I won't mind. Thanks for great work!

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -6,54 +6,54 @@ title:  Rust Documentation &middot; The Rust Programming Language
 # Rust Documentation
 
 If you haven't seen Rust at all yet, the first thing you should read
-is the introduction to the book, [**The Rust Programming
-Language**][book]. It will give you a good idea of what Rust is like,
+is the introduction to the book, [The Rust Programming
+Language][book]. It will give you a good idea of what Rust is like,
 show you how to install it, and explain its syntax and concepts. Upon
 completing the book, you'll be an intermediate Rust developer, and
 will have a good grasp of the fundamental ideas behind Rust.
 
 ## Learning Rust
 
-[**The Rust Programming Language**][book]. Also known as "The Book",
+[The Rust Programming Language][book]. Also known as "The Book",
 The Rust Programming Language is the most comprehensive resource for
 all topics related to Rust, and is the primary official document of
 the language.
 
-[**Rust by Example**][rbe]. A collection of self-contained Rust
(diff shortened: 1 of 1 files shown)


brson:
Thanks.

The reason they are bold is that I'm thinking of them not as links but as the titles of entries in a list - the bold is the style of the list.

This is also done on the [contributing page](https://www.rust-lang.org/contribute.html), so if changed in one place it should be in both. I rather like the way the contributing page looks though.

I definitely want to remove the bold from the first mention of the book on this page, to give the bold links more weight, like on the contributing page.

The documentation page looks good to me with or without the bold, but with my fondness for the contributing page, I'm on the fence. I'm going to sleep on it.

This might be solved by tweaking the font without removing the boldness. I don't notice fuzziness on Firefox on Windows.

cc @aturon @steveklabnik @AndrewBrinker @bjz you might have opinions on whether to use bold links on the doc and contributing page lists.

[Doc page without bold](http://brson.github.io/rust-www/documentation.html).

[Contrib page without bold](http://brson.github.io/rust-www/contribute.html).

AndrewBrinker:
Fira Sans has a bit of a beefy bold. There should be a nice middle ground of using `font-weight: 600` instead of the usual `font-weight: 700`. That way it's bolded, but not as harshly as it is now.

AndrewBrinker:
Huh. It doesn't look like the font is served at a semibold weight. I'll see if something else looks better.

bjz:
I agree with the removal of the use of a normal weight for small font sizes - this looks much easier to read, because the counters are more open. There is also enough hierarchy present in the different colour as opposed to the body text - making the links bold as well makes it seem a tad busy.

brson:
Thanks @listochkin @AndrewBrinker @bjz.

I'm merging this on its own merits and will think about #279 separately.

PR #278, opened 2016-01-16
Fixed chrome TOC numbering bug

AndrewBrinker:
This fixes the Chrome table of contents numbering bug. The solution was to add a new containing div that has the columns set on it. I also added `-webkit-perspective: 1;` to fix another Chrome bug when calculating container height.


alexcrichton:
Thanks!

PR #279, opened 2016-01-16
Removed bold from links to improve readability

AndrewBrinker:
Addresses #272.

The links have been changed to `h3` headers, with the explicit bolding removed. This appears nicer and avoids the typographic weirdness of bolding that some platforms display.


brson:
For now I prefer what we've got on the docs page, where each doc doesn't get a header and is not bold. I'll just merge https://github.com/rust-lang/rust-www/pull/302 to unbold the contrib page. Sorry about that.

AndrewBrinker:
No problem! Figured I'd put this up as a suggestion.

PR #280, opened 2016-01-16
Some FAQ fixes

Manishearth:


faq.md
```diff
@@ -117,7 +117,7 @@ There are several ways. You can:
 Why has Rust changed so much over time?
 </a></h3>
 
-Rust started with a goal of creating a safe but usable systems programming language. In pursuit of this goal it explored a lot of ideas, some of which it kept (lifetimes, traits) while others were discarded (the typestate system, green threading). Also, in the run up to 1.0 a lot of the standard library was rewritten as early designs were updated to best use Rust's features and provide quality, consistent cross-platform APIs.
+Rust started with a goal of creating a safe but usable systems programming language. In pursuit of this goal it explored a lot of ideas, some of which it kept (lifetimes, traits) while others were discarded (the typestate system, green threading). Also, in the run up to 1.0 a lot of the standard library was rewritten as early designs were updated to best use Rust's features and provide quality, consistent cross-platform APIs. Now that Rust has reached 1.0, the language is guaranteed to be "stable"; and while it may continue to evolve, code which works on current Rust should continue to work on future releases.
```
Manishearth: This section doesn't really address the implicit question of "will Rust keep changing?", and while it's answered in depth later I'd like to mention it here.

faq.md
```diff
@@ -356,6 +369,8 @@ How can I convert a <code>String</code> or <code>Vec&lt;T&gt;</code> to a slice
 Usually, you can pass a reference to a `String` or `Vec<T>` wherever a slice is expected.
 Using [Deref coercions](https://doc.rust-lang.org/stable/book/deref-coercions.html), [`String`s][String] and [`Vec`s][Vec] will automatically coerce to their respective slices when passed by reference with `&` or `& mut`.
 
+Methods implemented on `&str` and `&[T]` can be accessed directly on `String` and `Vec<T>`. For example, `some_string.char_at(0)` will work even though `char_at` is a method on `&str` and `some_string` is a `String`.
```
Manishearth: Another thing which IMO should be explicit, a lot of folks get confused by autoderef.

faq.md
```diff
@@ -971,7 +986,6 @@ What is the relationship between a module and a crate?
 - A module is a (possibly nested) unit of code organization inside a crate.
 - A crate contains an implicit, un-named top-level module.
 - Recursive definitions can span modules, but not crates.
-- Using another crate binds the root of its namespace into the user's namespace.
```
Manishearth: I have no idea what this means, we should be clearer or remove this. AIUI `use crate` and `use mod` work identically?

Manishearth:
cc @brson

faq.md
```diff
@@ -306,6 +306,19 @@ First, if every possibility is covered by the `match`, adding variants to the `e
...
+let x = match number {
+    1 | 2 | 3 => "small"
+    4 | 5 | 6 => "medium"
+    // does not handle all cases
+}
+// what is the value if `x` here if `number` was say, 10?
+```
+
+Such code leads to uninitialized values which may cause unsafety.
+
```
brson: The previous paragraph mentioned that a non-exhaustive match can panic, which would also work here. IOW, this is saying all cases *should* evaluate to a value, the other is saying some cases *could* panic. The two paragraphs sound like they are disagreeing.

faq.md
brson: Maybe the word "Using" is confusing. This is saying `extern crate` binds another namespace. cc @nikomatsakis you might care about this.

brson:
Thanks @Manishearth

PR #281, opened 2016-01-16
Clarify paragraph in "Is Rust garbage collected?" section of the FAQ

yberreby:
I [suggested rewording this part of the FAQ on reddit](https://www.reddit.com/r/rust/comments/40zxey/major_website_updates/cyyylwb), and @brson proposed that I make a PR, so here it is.

r? @brson


brson:
Thank you!

PR #287, opened 2016-01-21
Add ids to sections on downloads

hansihe:
Adding ids to the different sections on the download page, would allow linking directly to the relevant section when asking users to install a rust version.


alexcrichton:
Thanks!

PR #293, opened 2016-01-22
Update style of user groups page and rename to user-groups.html

brson:
I prefer hyphens over underscores for URLs and they are used
on the other pages on the site.

Leave a redirect.


brson:
cc @steveklabnik interested in your opinion on hypehns vs underscores in urls.

steveklabnik:
I am vaguely pro-hyphen but don't think that it's super important. I'm generally pro moving towards hyphens.

user_groups.html
```diff
@@ -0,0 +1,4 @@
+<DOCTYPE html>
+<html>
+  <meta http-equiv="refresh" content="0;url=user-groups.html"> 
+</html>
```
bjz: This has an underscored file name?

huonw: This is designed to be a redirect from the old underscored `rust-lang.org/user_groups.html` to the new hyphenated one, so that any links to it elsewhere don't rot.

alexcrichton:
:palm_tree:

PR #296, opened 2016-01-28
Get started on internationalization/localization

steveklabnik:
This is a work in progress PR for my proposal of https://internals.rust-lang.org/t/translations-for-rust/3126

Right now, there's just one commit: moving everything from the top level into `/en`. I've also put in redirects for the existing pages, to go to the English versions for now.

More to come!


community.md
```diff
@@ -1,210 +1,9 @@
...
-[contribute]: contribute.html
-[github]: https://github.com/rust-lang/rust
-[rfcs]: https://github.com/rust-lang/rfcs
-[team_reports]: https://github.com/rust-lang/subteams
-[issue_tracking]: https://github.com/rust-lang/rust/issues
+<html>
+  <head>
+    <meta http-equiv="refresh" content="0;/en/community.html" />
+  </head>
+</html>
```
huonw: Maybe this should be abstracted out into a layout, so making it handle the language redirects works properly, e.g. each file that needs to redirect is just

```markdown
---
layout: redirect
---
```

with _layouts/redirect.html containing:

```
<html>
  <head>
      <meta http-equiv="refresh" content="0;/en/{{ page.url }}" />
  </head>
</html>
```

(Untested)

huonw: Oh, it won't work with `user_groups`, so maybe the layout needs to be

```html
<html>
  <head>
      <meta http-equiv="refresh" content="0;/en/{% if page.redirect_to %}{{ page.redirect_to }}{% else %}{{ page.url }}{% endif %}" />
  </head>
</html>
```

and then `user_groups.md` is 

```markdown
---
layout: redirect
redirect_to: user-groups.html
---
```

(The others are unchanged.)

steveklabnik: I like this idea.

sanmai-NL:
One thing that needs to be correct from the start: not to use `en` but the specify the language variety. @steveklabnik thinks the current material is meant to be in American English, so please encode that. This is important to prevent mixups of varieties in contributed documentation. https://en.wikipedia.org/wiki/Language_localisation#Language_tags_and_codes

steveklabnik:
Ah yes, I was copying Ruby, but `en-US` would be better.

brson:
Is it really necessary to use `en-US` instead of `en` here? It's is more technically correct but: it will make the URLs uglier, and we don't *need* to ever add variations of the website for a single language, and probably won't.

AndrewBrinker:
It may be sufficient to use the ISO 639-1 codes rather than IETF language tags. On the one hand, the IETF tags are more exact, on the other the ISO 639-1 codes are better for URLs and may provide enough information. There's the question of how many translations we expect the site to have, as well.

steveklabnik:
We don't have to use `en-US`. It's just extra information, to signal that we're specifically using American English. I don't super super care, as I prefer the ergonomics of `en`, but the exact-ness of `en-US`.

brson:
It might be worth getting more input on the 'en' vs 'en-US' issue, though I imagine we can just make more redirects if we pick wrong.

This looks good to me with or without @huonw's suggested changes, though I agree factoring out the redirect seems better.

sanmai-NL:
1. You never know for sure what language varieties translations will be provided for in the future. Just as with code, it's easier to simplify later than to add edge cases later.
1. Whether the URL would be uglier is a matter of taste, thus it's subjective (unless someone has studied such usability aspects seriously?) and that makes it hazardous to assume some alternative URL is better. Many software projects, e.g. Mozilla's own Firefox, have `en-US` in their URL, even for end-user facing websites, e.g. https://www.mozilla.org/en-US/firefox/new/.
1. The point I made earlier is most important though: it is not clear that `en` means American English and in fact I've seen British English mixed up with American English on the website (e.g., ‘finalise’). This will at some point cause duplicate work when someone contributes documentation written in the wrong variety.
1. Believe it or not, such choices may aggravate non-American English speakers, when it appears that American is assumed to be the only English. Such things can be taken way too seriously and that will hurt inclusiveness. For instance, the R language has both `licence()` and `license()` and advertises both on startup of the REPL. I suppose some fighting went on to make it so ... Seems entirely unnecessary, but alas.

steveklabnik:
> such choices may aggravate non-American English speakers, when it appears that American is assumed to be the only English.

Oh I'd believe it.

steveklabnik:
Getting back to this, I've updated with @huonw 's suggestions, I think it's a lot cleaner. I've also done the `en-US`, as I think that I'm feeling like that's the right call, but we should discuss it.

There's one other little thing I've noticed while playing with this: the main header links link to `/documentation.html`, for example, which goes back to the root, which gets redirected. We may want to make these relative rather than absolute links?

alexcrichton:
I don't have much of an opinion on en vs en-US. I like en better because it looks nicer, but if en-US is "more correct" then it's more correct. I'd also recommend relative links wherever possible because you wouldn't necessarily want to click a link on the German translation and get redirected to the English translation by accident.

aturon:
+1 on en-us, that's definitely best practice here.

steveklabnik:
Whew!

I just completely re-did this PR, so that it's current with all changes to the website since back in January. I have set it up to do the `en-US` bit, as well as fixed the links in the headers so that they redirect to the same language, rather than back to the root URLs.

We should merge this sometime reasonably soon so it doesn't bitrot again.

steveklabnik:
https://github.com/rust-lang/rust-www/pull/433 has now been merged, so this is already out of date :9

alexcrichton:
cc @brson, think we're ready to pull the trigger on this?

steveklabnik:
Yeah, if you let me know when we're ready, I can make the final sync update and merge.

brson:
Yes, do it.

steveklabnik:
:confetti_ball:

PR #297, opened 2016-02-01
ENHANCE: Add Visual Studio Code plugin.

sanmai-NL:
The Visual Studio Code plugin works very well for me, while the Sublime Text 3 plugin is defunct. Therefore, this addition. Also, some language fixes, and an alphabetically reordered list of editors (to keep the peace in the lands of Editor Wars).

### DIFF
--- a/ides.html
+++ b/ides.html
@@ -44,15 +44,16 @@
 
 <p>Editor plugins:
 <br>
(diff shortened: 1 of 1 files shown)


brson:
Thanks!

PR #299, opened 2016-02-01
Added a syntax-highligting stylesheet for the FAQ

azerupi:
@brson  This should solve #246 but I can't test it locally because I am not able to start the jekyll server.

I didn't find a theme that matched the color scheme that the Ace editor uses on the front page so I based this theme on the one I use in mdBook, at least it would have the same colors as the new book..

If you don't like the theme, you can change / tweak it to match the theme of front page, however I don't have time to do that  so I guess `Some(syntax_highlighting)` is better than `None` :wink: 

Let me know if it works


brson:
Thanks! I only had to fix a typo. Looks lovely and should be deployed in 10 minutes or so.

PR #304, opened 2016-02-07
Simple typo fix

alevy:
"...other similar characteristic[s]"


brson:
Thanks for the PR.

I understand the intent here, but I think either could be correct, I feel the current wording reads better.

cc @rust-lang/core this is an update to the coc.

alevy:
Ok. I wasn't 100% sure either so figured I'd try. Closing...

PR #305, opened 2016-02-11
Clarify applicability of moderation to unofficial venues

Manishearth:
We were discussing this in #rust-community: A lot of Rust projects use the Rust CoC,
however the current CoC contains section about moderation/enforcement which
doesn't apply to these projects. This clarifies that section to avoid confusions
as to who to contact for enforcement.

cc @jdm @rust-lang/core @rust-lang/moderation @rust-lang/community

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -20,7 +20,8 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+
+These are the policies for upholding our community's standards of conduct. If you feel that a thread needs moderation, please contact the [Rust moderation team](/team.html#Moderation).
 
 1. Remarks that violate the Rust standards of conduct, including hateful, hurtful, oppressive, or exclusionary remarks, are not allowed. (Cursing is allowed, but never targeting another user, and never in a hateful manner.)
 2. Remarks that moderators find inappropriate, whether listed in the code of conduct or not, are also not allowed.
@@ -34,5 +35,7 @@ These are the policies for upholding our community's standards of conduct in our
 In the Rust community we strive to go the extra step to look out for each other. Don't just aim to be technically unimpeachable, try to be your best self. In particular, avoid flirting with offensive or sensitive issues, particularly if they're off-topic; this all too often leads to unnecessary fights, hurt feelings, and damaged trust; worse, it can drive people away from the community entirely.
 
 And if someone takes issue with something you said or did, resist the urge to be defensive. Just stop doing what it was they complained about and apologize. Even if you feel you were misinterpreted or unfairly accused, chances are good there was something you could've communicated better — remember that it's your responsibility to make your fellow Rustaceans comfortable. Everyone wants to get along and we are all here first and foremost because we want to talk about cool technology. You will find that people will be eager to assume good intent and forgive as long as you earn their trust.
+
+The enforcement policies listed above apply to all official Rust venues; including official IRC channels (#rust, #rust-internals, #rust-tools, #rust-libs, #rustc, #rust-beginners, #rust-docs, #rust-community, #rust-lang, and #cargo); GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org (users.rust-lang.org, internals.rust-lang.org). For other projects adopting the Rust Code of Conduct, please contact the maintainers of those projects for enforcement. If you wish to use this code of conduct for your own project, consider explicitly mentioning your moderation policy or making a copy with your own moderation policy so as to avoid confusion.
 
 *Adapted from the [Node.js Policy on Trolling](http://blog.izs.me/post/30036893703/policy-on-trolling) as well as the [Contributor Covenant v1.3.0](http://contributor-covenant.org/version/1/3/0/).*


Manishearth:
er, cc @bstrie @erickt @skade  since the community ping doesn't work.

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+_The enforcement policies listed below apply to all official Rust venues; including official IRC channels; GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org. For other projects adopting the Rust CoC, please contact the maintainers for enforcement. Feel free to contact the Rust moderation team if the need arises, however._
```
bstrie: > Feel free to contact the Rust moderation team if the need arises, however.

Feels vague. If the need arises for what?

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
...
+
+
+These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels. If you feel that a thread needs moderation, please contact the [Rust moderation team](/team.html#Moderation).
```
bstrie: This chunk seems to have some redundancy with the new chunk above.

steveklabnik:
general :+1:

Manishearth:
> Feel free to contact the Rust moderation team if the need arises, however.

So, the reason I put this in here was as an escape hatch for when the CoC violation is from a maintainer. However, you're right, this is a rather strange way of wording it. I'm not sure how to word it otherwise; any suggestions?

conduct.md
aturon: Agreed, I think that last sentence should be dropped.

aturon:
@Manishearth Can you lay out the scenario you envision in more detail? You seem to be saying that, if some external project's maintainer is violating the CoC within their own project, the Rust moderation team has a role to play there -- but I don't see what that role is.

aturon:
(This last issue aside, the overall clarification here is very good.)

alexcrichton:
Seems like a worthwhile clarification to me as well, I agree with @bstrie and @aturon that the part about contacting the Rust moderation team if this is in use elsewhere can probably be elided for now.

Manishearth:
Removed that bit.

So the current wording encourages "contact the rust mod team for any issue on any project"  (for someone unfamiliar with the larger community who reads conduct.html after following a link in a README in an unofficial project). The primary clarification in this PR fixes that; since we don't want situations where someone contacts the mod team for something outside their hands (and we reply  "sorry, can't do anything" -- it's an overall bad user experience). However, in that process we lose the ability to be notified of cases when a maintainer of a project adopting the CoC makes an egregious violation. In such a case it's very likely that the project adopted the CoC just to tick a box, and it would be nice to see if we can provide an alternate venue for this to be handled. Not that we can enforce things on random projects, but we can perhaps open a dialogue with the maintainer or something.

Anyway, it's an unlikely edge case, which we probably don't need to worry about.

aturon:
Yeah, I'd hope that if another project copies this updated text, they'd also change the list of applicable venues and contact point for moderation.

aturon:
Added to core team meeting agenda for next week, though if everybody gives a +1 before that we can potentially merge earlier.

huonw:
+1 from me.

jdm:
>Yeah, I'd hope that if another project copies this updated text, they'd also change the list of applicable venues and contact point for moderation.

The reason I originally brought this up is that many projects have text like "We follow the Rust Code of Conduct" and link directly to the original text. This is easy to do (while feeling like you're making a smart choice) and should really be an anti-pattern, in my opinion.

aturon:
@jdm Ahha, I didn't think about that (and see that I read @Manishearth's earlier comment too quickly). Tricky!

While I understand @Manishearth's point that the mod team might be able to help with trouble in an external repo, if the premise is that the project maintainer of that repo is not enforcing -- or is perhaps violating -- the CoC, it's hard to see that ending well. I guess my feeling is, as much as I'd like our CoC to be a standard for the whole community (and not just our venues), it's not realistic for us to actually *enforce* it elsewhere in any real sense, and I think we should avoid offering to get involved.

For the case where the CoC is just linked: I wonder if we can add some text at the bottom, where we currently give attribution, and explicitly ask that other projects who wish to adopt this CoC actually copy it and adjust venues/contact points appropriately, rather than linking. We can also point out that our CoC is updated from time to time, and downstream projects should adopt such changes consciously, rather than just by virtue of linking to ours.

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+_The enforcement policies listed below apply to all official Rust venues; including official IRC channels; GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org. For other projects adopting the Rust Code of Conduct, please contact the maintainers for enforcement._
```
SimonSapin: Which of the many Rust-related IRC channels are official? http://edunham.net/2015/07/31/how_many_rust_channels_are_there.html

pnkfelix: realistically the rust-mods team can only assist with the channels where we have op privileges.  that's a small enough set that we might just list them explicitly here...

skade: @pnkfelix that would play down the role of moderation. They can only _ban_, but they can discuss and recommend solutions.

aturon: As I said elsewhere, I think it's important that our CoC/moderation team only be applied to venues that are officially part of the Rust community and have agreed to abide by the CoC.  Listing the IRC channels that fall into that category seems good to me.

Manishearth:
> The reason I originally brought this up is that many projects have text like "We follow the Rust Code of Conduct" and link directly to the original text. This is easy to do (while feeling like you're making a smart choice) and should really be an anti-pattern, in my opinion.

IMO one of the reasons for doing this is that you get any improvements to the core CoC for free.

But yeah, perhaps we should put a note at the bottom. How about this?

> _If you wish to adapt this code of conduct for your own project, consider explicitly mentioning your moderation policy or making a copy with your own moderation policy so as to avoid confusion_

aturon:
> IMO one of the reasons for doing this is that you get any improvements to the core CoC for free.

I have mixed feelings about this -- it's sort of akin to one of those agreements where one clause is "and the other party can change the agreement at any time without notification."

I do realize that not every project feels qualified to judge CoC changes themselves -- and indeed, we recently pulled in some "best practice" changes from another source. But I'm not sure that the Rust project is really the right place to centralize these concerns -- better to link to a standalone CoC designed for that purpose, I'd think.

Manishearth:
Done.

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+_The enforcement policies listed below apply to all official Rust venues; including official IRC channels (#rust, #rust-tools, #rust-libs, #rustc, #rust-beginners, #rust-docs, #rust-community, #rust-lang); GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org. For other projects adopting the Rust Code of Conduct, please contact the maintainers for enforcement._
```
aturon: Missing rust-internals on the IRC list. Also "all forums under rust-lang.org" may as well spell out users and internals, given that all the other lists are expanded.

Manishearth: updated

skade:
This shouldn't hold back this issue, but I've got a couple of comments on this: I think adopting a "foreign" Code of Conduct instead of writing your own is a good and a bad thing. Good: you benefit from the work of others. You should still spend time to _understand and interpret_ that work, but writing CoCs yourself without being fully aware on how they work is problematic. Bad: you run into the danger of just signing a piece of paper and to skip the awareness process. Also, you might adopt something which doesn't quite fit your use case.

For example, I usually recommend user groups to be run "in the faith of the Rust Code of Conduct and using the $you_favourite_coc". I obviously usually recommend http://berlincodeofconduct.org/, but I'm obviously biased.

Maybe adding a statement about how to (not) adopt the CoC would help there.

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+_The enforcement policies listed below apply to all official Rust venues; including official IRC channels (#rust, #rust-internals, #rust-tools, #rust-libs, #rustc, #rust-beginners, #rust-docs, #rust-community, #rust-lang, and #cargo); GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org (users.rust-lang.org, internals.rust-lang.org). For other projects adopting the Rust Code of Conduct, please contact the maintainers for enforcement._
```
nikomatsakis: > For other projects adopting the Rust Code of Conduct, please contact the maintainers for enforcement.

I found this confusing. At first I thought you meant the "Rust maintainers". I think you should say:

> For other projects adopting the Rust code of conduct, please contact the maintainers of those projects for enforcement.

Though the sentence feels a bit awkward to me (but I can't find a better way to phrase it just now).

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
...
+
+
+These are the policies for upholding our community's standards of conduct. If you feel that a thread needs moderation, please contact the [Rust moderation team](/team.html#Moderation).
```
nikomatsakis: I sort of think that this paragraph should still come first? The clarification above just seems to provide more detail about what kinds of threads and venues are "covered".

Manishearth: done

conduct.md
Manishearth: done

conduct.md
```diff
@@ -20,7 +20,10 @@ title: The Rust Code of Conduct &middot; The Rust Programming Language
 
 ## Moderation
 
-These are the policies for upholding our community's standards of conduct in our communication channels, most notably in Rust-related IRC channels.
+
+These are the policies for upholding our community's standards of conduct. If you feel that a thread needs moderation, please contact the [Rust moderation team](/team.html#Moderation).
+
+_The enforcement policies listed below apply to all official Rust venues; including official IRC channels (#rust, #rust-internals, #rust-tools, #rust-libs, #rustc, #rust-beginners, #rust-docs, #rust-community, #rust-lang, and #cargo); GitHub repositories under rust-lang, rust-lang-nursery, and rust-lang-deprecated; and all forums under rust-lang.org (users.rust-lang.org, internals.rust-lang.org). For other projects adopting the Rust Code of Conduct, please contact the maintainers of those projects for enforcement._
```
aturon: I think it's worth moving this paragraph to below the other details about moderation, which is more important.

brson:
I'm happy with this generally, except that I do not like that there are *3* italicized paragraphs in this section, which reduces their impact.

aturon:
**Procedural note**: the core team discussed this PR today, and everyone is on board with the basic change, modulo a few minor textual edits (which have been left in comments).

Manishearth:
Addressed. I moved that paragraph to the bottom, merging it with the other one, and removed italics (unnecessary now that it's at the bottom)

aturon:
Thanks @Manishearth!

PR #312, opened 2016-03-02
Add intellij to short list

gmorenz:
Just noticed that this was missing from the first list of plugins. Since we expand on it below it seems like it should be here as well.


aturon:
Thanks!

PR #316, opened 2016-03-04
Add myself to the list of compiler team members

nagisa:
Note: I didn’t test the changes, so it might be possible I broke something.

### DIFF
--- a/team.md
+++ b/team.md
@@ -46,6 +46,9 @@ people:
   michaelwoerister:
     name: Michael Woerister
     irc: mw
+  nagisa:
+    name: Simonas Kazlauskas
+    irc: nagisa
   niconii:
     name: Nicolette Verlinden
     irc: niconii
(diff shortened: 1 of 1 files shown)


nikomatsakis:
@nagisa There might be some confusion. You're a committer for the compiler, but you're actually not a member of the compiler subteam (yet). This is not in any way to minimize the work you've been doing, which is great. It's just the way the process works: new members are added when the team decides that, for some time, someone has been making major contributions to the design and overall architecture of the compiler. In any case, the role of the subteam is (by design) fairly limited: most all discussion takes place on internals/IRC and is open to all, but members make the final decisions on RFCs as well as performing more mundane tasks like triage. Anyway, I'm going to close this PR; let me know if anything's unclear.

nagisa:
Aw right, my bad, things are confusing.

nikomatsakis:
@nagisa no worries! :)

PR #320, opened 2016-03-13
update link to the cross-compilation guide

matklad:
The page for the old link says that it has been deprecated in favor of the new link:

>Did Google bring you here? Anyways, I'd like to inform you that this document is DEPRECATED, that doesn't mean that this information is wrong, but it could be; it means that this document won't receive updates or fixes in the case it has wrong information.


alexcrichton:
Thanks!

PR #321, opened 2016-03-16
Clean up JavaScript

tillarnold:
In the JavaScript files there were mixed spaces and tabs.  I also simplified the detect_platform function and put it into the IIFE.


alexcrichton:
Thanks!

PR #332, opened 2016-03-30
Add alt attribute to images for team members

fbergr:
Images should have an alt attribute, except under certain
conditions.

[Requirements for providing text to act as an alternative for images](https://www.w3.org/TR/html5/embedded-content-0.html#alt)


alexcrichton:
Neat!

PR #334, opened 2016-03-31
Update VS instructions to link to Microsoft Build Tools 2015

brson:
cc @retep998

### DIFF
--- a/downloads.html
+++ b/downloads.html
@@ -183,9 +183,13 @@ title: Downloads &middot; The Rust Programming Language
 	  use the GNU build.
 	</p>
 	<p>
-	  MSVC builds of Rust additionally require an
-	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio 2013 (or later)</a>
-	  so rustc can use its linker. Make sure to check the "C++ tools" option.
+	  MSVC builds of Rust additionally require the Microsoft Visual
+          C++ build tools for Visual Studio 2013 or later.
+	  The easiest way to acquire the build tools is by installing
+	  <a href="http://go.microsoft.com/fwlink/?LinkId=691126">Microsoft Visual C++ Build Tools 2015</a>
+	  which provides just the Visual C++ build tools. Alternately, you can
+	  <a href="https://www.visualstudio.com/downloads/">install</a>
+	  Visual Studio 2015 or Visual Studio 2013 and during install select the "C++ tools".
 	  No additional software installation is necessary for basic use of the GNU build.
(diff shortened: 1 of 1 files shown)


downloads.html
```diff
@@ -183,9 +183,13 @@ <h2 id="nightly">Nightly&nbsp; ({{ site.nightly }})</h2>
 	  use the GNU build.
 	</p>
 	<p>
-	  MSVC builds of Rust additionally require an
-	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio 2013 (or later)</a>
-	  so rustc can use its linker. Make sure to check the "C++ tools" option.
+	  MSVC builds of Rust additionally require the Microsoft build
+	  tools for Visual Studio 2013 or later so rustc can use the
+	  Visual C++ linker. These are available either as a
+	  <a href="https://www.microsoft.com/en-us/download/details.aspx?id=48159">small download without Visual Studio</a>,
```
petrochenkov: Nah, "Microsoft Build Tools" and "Microsoft Visual C++ Build Tools" are different things, this is a link for the former and we need the latter. They can be downloaded from https://www.visualstudio.com/downloads/download-visual-studio-vs -> Tools for Visual Studio 2015 -> Microsoft Visual C++ Build Tools 2015
(I suppose they'll appear on microsoft.com/en-us/download too eventually.)

downloads.html
```diff
@@ -183,9 +183,13 @@ <h2 id="nightly">Nightly&nbsp; ({{ site.nightly }})</h2>
 	  use the GNU build.
 	</p>
 	<p>
-	  MSVC builds of Rust additionally require an
-	  <a href="https://www.visualstudio.com/downloads">installation of Visual Studio 2013 (or later)</a>
-	  so rustc can use its linker. Make sure to check the "C++ tools" option.
+	  MSVC builds of Rust additionally require the Microsoft build
+	  tools for Visual Studio 2013 or later so rustc can use the
```
petrochenkov: Nit: VC++ Build Tools for VS < 2015 don't exist

brson: The way I was thinking of it the "Microsoft build tools" do exist in VS 2013 - they are the tools for building that have always been distributed with VS. It's the product "Microsoft Build Tools" that didn't exist.

downloads.html
brson: Wow, I really don't want to give people directions on how to click to that download. There must be a  more direct link....

brson:
I've updated it to link directly to the installer.

PR #336, opened 2016-04-01
Improve meta tags according Bootstrap documentation

fbergr:
Bootstrap framework is not supported in the old Internet Explorer
compatibility modes. Bootstrap recommends usage of following meta
tag to be sure that the latest rendering mode for IE is used:

`<meta http-equiv="X-UA-Compatible" content="IE=edge">`

According Bootstrap documentation first three meta tags should be:

```
<meta charset="...">
<meta http-equiv="...">
<meta name="viewport" content="...">
```

See: http://getbootstrap.com/getting-started/

### DIFF
--- a/_layouts/basic.html
+++ b/_layouts/basic.html
@@ -3,8 +3,9 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <title>{{ page.title }}</title>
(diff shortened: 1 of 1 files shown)


alexcrichton:
Sure!

Issue #223, opened 2015-11-19
Code on the front page is not "editable and runnable" if JavaScript is disabled


Zarathustra30:
While this makes perfect sense, it turns the statement "This code is editable and runnable!" into a lie. A modified version without that claim should served instead.

AndrewBrinker:
Looks like this issue has been fixed, and can be closed. @brson?

brson:
Yep!

PR #337, opened 2016-04-06
Hide editable code comment when JS is disabled

CalMlynarczyk:
See #223.

### DIFF
--- a/_includes/example.rs.html
+++ b/_includes/example.rs.html
@@ -1,4 +1,4 @@
-<pre class='rust'>// This code is editable and runnable!
+<pre class='rust'>
 <span class='kw'>fn</span> main() {
 <span class='comment'>// A simple integer calculator:
 // `+` or `-` means add or subtract by 1


alexcrichton:
Thanks for the PR! Could this comment actually get added if JS is enabled, however?

CalMlynarczyk:
@alexcrichton [_editor.js_](https://github.com/rust-lang/rust-www/blob/77e65109c281aa56216a8967136f495170825227/_includes/editor.js) hides the static element on load. It is in the page DOM, but it does not get displayed to the user. The code in the real editor uses the separate _exampls.rs_ file, which has the comment at the top. It looks like this got changed back in PR #39 for some [supposed] very minor performance reasons, but based on #223, it was incorrect.

**editor.js : 13**

    // Fetching DOM items
    var activeCode = document.getElementById("active-code");
    var staticCode = document.getElementById("static-code");
    
    /* Other element logic... */
    
    // JS exists, display ACE editor
    staticCode.style.display = "none";
    activeCode.style.display = "block";

alexcrichton:
It looks like the same `example.rs.html` template here gets loaded in both situations though, right? In that case if JS is enabled we'll display the "fancier UI" with buttons, but we still won't have the comment indicating that this code is runnable?

CalMlynarczyk:
If you look in [_index.html_](https://github.com/rust-lang/rust-www/blob/309526268a13e0e2d8b04186e47d9cbf97b931a0/index.html), there are two separate elements: `active-code` and `static-code`. `active-code` contains the `editor` element which loads the [_example.rs_](https://github.com/rust-lang/rust-www/blob/bb4c538db3c6809eaac3ede6680fb380d57503c9/_includes/example.rs) file. This contains the code _with_ the comment included at the top for when the editor is loaded by JS.

The `static-code` element loads the content in [_example.rs.html_](https://github.com/rust-lang/rust-www/blob/299cb007c007f5883ba7bda8011545a9ddef9bc2/_includes/example.rs.html), which my change takes out that top comment, because it does not make sense in its static context when JS is disabled.

alexcrichton:
Wow I'm clearly not reading this closely enough... Sorry for the confusion!

PR #346, opened 2016-04-28
Add a "production users" page and give it some initial content

brson:
[Rendered](http://brson.github.io/rust-www/friends.html).

This page showcases organizations using Rust in production. It's useful and expected marketing for both us and those listed. The page is presently linked from a new FAQ entry, and in the future will be linked more prominently.

The page links to a pre-filled form on the www issue tracker that asks for all the required information. Every logo we add needs to come with contact information in case we need to make substantive changes to the page in the future. To submit the contact info they can either put it in the form or submit to `user-logos@rust-lang.org` which is an alias for the core team. The contacts are maintained in a private spreadsheet to which the core team has access.

Once this is live I will post to the forum and reddit to try to encourage more users to contribute their logos.

All logos approved by their respective owners.

cc @jonathandturner @rust-lang/core


brson:
I'm interested in what you think of "and that makes us happy". There are very few words on this page, so it's important that they all be good.

brson:
Another thing to note: similar pages put the info about how to submit your logo at the bottom of the page, which I think makes us seem less greedy. I have a concern right now that the text is too us-focused instead of them-focused.

steveklabnik:
Big :+1: overall!

> I'm interested in what you think of "and that makes us happy".

and

> I have a concern right now that the text is too us-focused instead of them-focused.

I think I share this slightly. I found "and that makes us happy" confusing to read; I thought it was about how we made _them_ happy, and that there would be specific reasons there. I do like the sentiment, but I will try to think of an alternate wording. First shot:

> We love people doing wonderful things with Rust! These organizations use Rust in production. To add...

Not sure if that's too heavy-handed...

jonathandturner:
+1 on putting the text at the bottom

I like how Ember (http://emberjs.com/ember-users/) just has a brief bit about the users and how they might have different usage.  TypeScript (http://www.typescriptlang.org/community/friends.html) just doesn't say anything but "Friends of TypeScript" which also works.

brson:
I wonder if asking for issues to submit logos risks us not noticing the issues. I know I periodically check the rust-www PRs, but not often the issue tracker.

steveklabnik:
I have emails turned on for this repository, so I will get an email about it, atl east.

brson:
I pushed an update that changes the top to say only "Who uses Rust?", and puts new instructions at the bottom.

nrc:
Nice!

I wonder if we should be more explicit about this being production users? That info only seems to be hinted at in the user-facing text at the moment. If we want to be more general and this be about significant users of Rust, then maybe we should add the Universities which use it in courses and research projects, etc.

brson:
@nrc It needs to be both clear and subtle that this isn't a list for *every user*, and the definition of "production" needs to be kinda fuzzy.

Right the hint is in the footer: "If your organization uses Rust in production". @jonathandturner suggested to me that the title could be a bit more delicate than "Who uses Rust?". TypeScript e.g. says "Friends of Rust", but that is too inviting to go without further qualification. Everybody's a friend of Rust!

Another possibility taken from @jonathandturner for what to put in this page: "Friends of Rust - Organizations betting on Rust in production".

brson:
I like the idea of adding Universities.

nrc:
> It needs to be both clear and subtle that this isn't a list for every user

Yeah, it's kind of delicate, I don't want to list every weekend project written in Rust

> Organizations betting on Rust in production

I love this definition of significance

brson:
OK. Further updates. Now the url is `friends.html` and the header is "Friends of Rust (Organizations running Rust in production)".

I am happy with this and intend to publish it very shortly.

skade:
I'm late to the party, but how is tessel on that page? Their rust support is dropped and basically a community effort currently.

PR #361, opened 2016-05-03
Using PNG instead of SVG for PingCAP Logo

c4pt0r:
Some browsers(e.g. Chrome for Windows) do not support the font in
original svg, so use png instead.

r? @brson


brson:
Thanks!

PR #363, opened 2016-05-03
Correct information about IntelliJ plugin

kryptan:
It is written in Kotlin, not in Java.


steveklabnik:
Ah, nice catch. Thanks!

PR #366, opened 2016-05-03
Fix "represented" typo

mitchhentges:
Other than the odd typo, these docs are great! Thanks guys!


steveklabnik:
Thanks so much!

PR #372, opened 2016-05-08
Update info about IntelliJ-Rust plugin

matklad:
Hi!

We've moved the repo to the dedicated organization, and while the google group is basically dead, the gitter is rather active.


aturon:
Thanks!

Issue #381, opened 2016-05-17
"Rust Development" header on community page isn't rendering properly


Frijol:
![screen shot 2016-05-17 at 9 14 58 am](https://cloud.githubusercontent.com/assets/454690/15325439/08eca9e8-1c10-11e6-99a5-a39d1e69d0a9.png)

As you see, the markdown `##` isn't turning "Rust Development" into a header.

Looks like it's because of the `<a>` tag above it (none of the other headers have this), but not PR'ing its removal in case you're using that for something.
![screen shot 2016-05-17 at 9 17 07 am](https://cloud.githubusercontent.com/assets/454690/15325520/48c6760c-1c10-11e6-8b6c-88929d9cab96.png)

Happy to PR if you can confirm you don't need that line for anything!

bjz:
I'm guessing it was there to create a reference to the header. But that should be automatically inserted as an id by markdown, and indeed the other headers on that page have ids generated for them.

steveklabnik:
Not all markdown renderers give you this ID, including when you view a MD
file on github directly. but consistency per file is best.

PR #382, opened 2016-05-17
fixes https://github.com/rust-lang/rust-www/issues/381

Frijol:
Hopefully should unbreak the rendering of this header on the site


steveklabnik:
Thank you!

Issue #284, opened 2016-01-17
Provide some context for the terminology used when describing HKTs


boblehest:
When reading the [FAQ entry about HKTs](https://github.com/rust-lang/rust-www/blob/master/faq.md#what-are-higher-kinded-types), I'd like to know where the terminology stems from.

Example: 

> Any complete type, like i32, bool, or char is of kind *.
> A type with one parameter, like Vec<T> is of kind * -> *

Does this stem from another language? Is it from a mathematics field?

AndrewBrinker:
This is all from the field of type theory. The notation is included as I felt it provides a clearer picture of what higher-kinded types are. They are, in essence, "types with holes". That is, they have unfilled type parameters. The notation is designed to show both the relationship between the number of unfilled parameters and the type's "kindedness" as well as hint at the dual nature of a higher-minded type as both a type itself and as a function that returns another type, possibly also with holes.

From a logic perspective, kinds are "one level up" from types. That is, they are similar to the simply-typed lambda calculus, with the type `*` (called "type") which I call a "complete type" (type without holes/unfilled parameters) in the answer.

This answer was one of the toughest to write, and I'm happy to take another crack at clarifying it.

boblehest:
I think the answer is well written, I just think it would be pertinent to tell the readers what branch of mathematics they're currently exploring, in case they want to read more about it. :)

"In type theory, any complete type [..] is denoted by a *", or something like that.

brson:
I believe this is fixed by the most recent changes.

Issue #262, opened 2016-01-15
Misnomer in FAQ.


Aaronepower:
In the [FAQ](https://github.com/rust-lang/rust-www/blob/master/faq.md#what-is-the-difference-between-consuming-and-moving) there is a section with the following heading, and body text. However to my knowledge passing by value doesn't mean the same thing as the rest. Passing by value is equivalent of the `Copy` trait, where the original owner can still use the data after it has passed the value to a different owner.
***
### What is the difference between passing by value, consuming, moving, and transferring ownership?

These are different terms for the same thing. In all cases, it means the value has been moved to another owner, and moved out of the possession of the original owner, who can no longer use it.
***

steveklabnik:
We tend to use them in the same way in Rust-world, regardless if things are Copy or not. Other languages don't have the ownership rules, so they don't do the kind of invalidation we do.

Aaronepower:
Yes, but it is an FAQ for people who aren't familiar with Rust. There should be some clarity that the term is being used in a different way than someone coming from a different language might expect. This could be clarified in the section following which does discuss the `Copy` trait.

AndrewBrinker:
I think you're right. It's probably worthwhile to mention the `Copy` trait and state that a move does occur if the passed-in type implements it, and make it clear that both the move and copy versions are considered "pass by value" (the notion of whether ownership was moved or copied is opaque from the perspective of the function).

Issue #248, opened 2016-01-14
More FAQs


brson:
There are still FAQs not answered:

* "What does the name 'Rust' mean?"
* "Why does a tiny program like "Hello World" produce such a large executable file?" - [answer](https://users.rust-lang.org/t/why-do-rust-programs-use-more-memory-than-the-c-haskell-and-ocaml-versions/3224/1).

antoyo:
[The second question](https://users.rust-lang.org/t/why-do-rust-programs-use-more-memory-than-the-c-haskell-and-ocaml-versions/3224/1) is about RAM usage, not executable size.

brson:
Comparison to swift would be useful.

Manishearth:
@antoyo Yes, but the additional RAM usage is _caused_ by a larger executable (which is loaded into RAM).

For swift I guess something along the lines of it having a very similar type system without the ownership/borrowing, and using automatic reference counting for memory safety?

antoyo:
@Manishearth In the thread on rust-lang, I learned that using the system allocator lower the RAM usage. Does this mean that jemalloc is using 10 MB of RAM?

Manishearth:
That could be the case, but I doubt it; it probably means you're statically linking to jemalloc which increases the binary size (whereas the system allocator exists on your machine anyway)

brson:
"What's the point of inner attributes?" / "Why are there so many attribute forms?" / "Why are doc comments attributes?"

petrochenkov:
"Why does my Windows application show console?"

kud1ing:
"Which string type should i use?"

http://hermanradtke.com/2015/05/03/string-vs-str-in-rust-functions.html

kud1ing:
"Can i accept both `&str` and `String`?"

http://hermanradtke.com/2015/05/06/creating-a-rust-function-that-accepts-string-or-str.html

F001:
Why there is no incremental/decremental operators ( ++/-- ) in Rust? 

Well, I know these operators lead to troubles, but I can't find the canonical explanation.

AndrewBrinker:
I am happy to take a crack at answers to the above-listed questions. I can probably whip up reasonable answers to all of them but the naming question. If anyone familiar with the history has a good explanation for the name, I'm happy to write up an answer that explains it.

PR #394, opened 2016-06-01
Additional FAQ answers

AndrewBrinker:
I've added answers for a number of questions, including some mentioned in #248, #262, and #284. There are still some questions left to be answered, but I figure this is a good start.

r? @brson


brson:
OMG awesome.

faq.md
```diff
@@ -144,6 +144,16 @@ No, you cannot. Rust works hard to provide strong guarantees about the stability
...
 Things stabilize all the time, and the beta and stable channels update every six weeks. If you're waiting for a feature to be available without using the nightly channel, you can locate its tracking issue by checking the [`B-unstable`](https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aopen+tracking+label%3AB-unstable) tag on the issue tracker.
 
+<h3><a href="#what-are-feature-gates" name="what-are-feature-gates">
+What are "Feature Gates"?
+</a></h3>
+
+"Feature gates" are attributes that enable the usage of unstable language
+features. They can only be activated on the nightly channel, and allow
+developers to opt into experimental features before they are stabilized.
+Feature gates follow the same stabilization procedure as the standard library.
```
brson: "nightly channel" -> "nightly release channel" please.

I think this is a little unclear because it implies that only language features are feature-gated, and "follow the same stabilization procedure as the standard library". But both language and library features are feature-gated. Feature-gating is the rust stabilization mechanism.

Some alternate ideas for text:

> "Feature gates" are the mechanism Rust uses to stabilize features of the compiler, language, and standard library. A feature that is "gated" is accessible only on the nightly release channel, and then only when it has been explicitly enabled through `#[feature]` attributes or the `-C unstable-options` command line argument. When a feature is stabilized it becomes available on the stable release channel, and does not need to be explicitly enabled. At that point the features is considered "ungated". Feature gates allow developers to test experimental features while they are under development, before they are available in the stable language.

faq.md
```diff
@@ -351,6 +361,13 @@ How can I convert between numeric types?
 
 There are two ways: the `as` keyword, which does simple casting for primitive types, and the [`Into`][Into] and [`From`][From] traits, which are implemented for a number of type conversions (and which you can implement for your own types). The [`Into`][Into] and [`From`][From] traits are only implemented in cases where conversions are lossless, so for example, `f64::from(0f32)` will compile while `f32::from(0f64)` will not. On the other hand, `as` will convert between any two primitive types, truncating values as necessary.
 
+
+<h3><a href="#why-doesnt-rust-have-increment-and-decrement-operators" name="why-doesnt-rust-have-increment-and-decrement-operators">
+Why doesn't Rust have increment and decrement operators?
+</a></h3>
+
+Rust prefers to avoid single purpose features like the increment and decrement operators, `++i` and `--i`. In addition, debate other whether to use the prefix or postfix forms of the operators, and potential semantic differences between them if both are provided, is more of a headache than is acceptable for such a limited feature.
+
```
brson: I wonder what others think of this explanation. Honestly, I've never put much thought into it, but it's a pretty minor feature with known semantic weirdness.

I wonder if we can word this so it doesn't personify Rust, "Rust prefers".

I'm not sure how true it is that we avoid "single purpose features". In fact if you think of "single purpose" as "orthogonal", then Rust *strives* for single-purpose features - we like small features that do one thing and work cleanly with other features.

This probably deserves a link to explanations of problems with the prefix increment operator.

brson: Here's an old  faq an the subject: https://github.com/rust-lang/rust/pull/23186/files

faq.md
```diff
@@ -415,6 +432,42 @@ The [`char`][char] type is UTF-32. If you are sure you need to do a codepoint-at
 
 For a more in-depth explanation of why UTF-8 is usually preferable over UTF-16 or UTF-32, read the [UTF-8 Everywhere manifesto](http://utf8everywhere.org/).
 
+<h3><a href="#why-are-there-multiple-types-of-strings" name="why-are-there-multiple-types-of-strings">
```
brson: This anchor has the wrong name, which is also duplicated below.

faq.md
```diff
@@ -415,6 +432,42 @@ The [`char`][char] type is UTF-32. If you are sure you need to do a codepoint-at
...
+If you're working with C code through the Rust FFI, use the C-compatible string types. If you're working with the operating system, including command line arguments, use the OS-compatible string types. If you're working with paths, use the system path string types. Otherwise, use the UTF-8 string types.
+
+<h3><a href="#why-are-there-multiple-types-of-strings" name="why-are-there-multiple-types-of-strings">
+Why are there multiple types of strings?
+</a></h3>
+
+Rust's difference string types serve different purposes. `String` and `str` are UTF-8 encoded general-purpose strings. `OsString` and `OsStr` are encoded according to the current platform, and are used when interacting with the operating system. `CString` and `CStr` are the Rust equivalent of strings in C, and are used in FFI code, and `PathBuf` and `Path` are convenience wrappers around `OsString` and `OsStr`, providing methods specific to path manipulation.
+
+
+<h3><a href="#why-are-there-multiple-types-of-strings" name="why-are-there-multiple-types-of-strings">
```
brson: Wrong anchor.

faq.md
AndrewBrinker: Yup. You're right. Copy and paste mistake.

faq.md
AndrewBrinker: Hmm... that's a better FAQ. I may just reuse it.

faq.md
AndrewBrinker: That text is better. Thanks!

faq.md
```diff
@@ -415,6 +432,42 @@ The [`char`][char] type is UTF-32. If you are sure you need to do a codepoint-at
...
+| System path   | `Path`       | `PathBuf`    |
+
+If you're working with C code through the Rust FFI, use the C-compatible string types. If you're working with the operating system, including command line arguments, use the OS-compatible string types. If you're working with paths, use the system path string types. Otherwise, use the UTF-8 string types.
+
+<h3><a href="#why-are-there-multiple-types-of-strings" name="why-are-there-multiple-types-of-strings">
+Why are there multiple types of strings?
+</a></h3>
+
+Rust's difference string types serve different purposes. `String` and `str` are UTF-8 encoded general-purpose strings. `OsString` and `OsStr` are encoded according to the current platform, and are used when interacting with the operating system. `CString` and `CStr` are the Rust equivalent of strings in C, and are used in FFI code, and `PathBuf` and `Path` are convenience wrappers around `OsString` and `OsStr`, providing methods specific to path manipulation.
+
```
brson: I think these two questions, 'what string should i use', and 'why are there multiple types of string' can be combined, since they both cover similar territory about the 4 string pairs.

This can also be expanded quite a bit. As it reads it doesn't lead you to the obvious answer first, which is "use &str". I'd say this explanation should start by explaining when to use the slice type vs when to use the owned type, which is itself quite a complicated topic. Slice type for arguments; slice type for return types when you are returning an interior pointer; String types for return types when the string is being generated by the function; etc.

Only then unveil the 4 string pairs and explain when each of them is necessary.

The linked article seems to lay it out in way I'd expect. I wonder if it can be outlined and condensed here. Though since the article is so good it may be worth not trying to explain *too much* here.

faq.md
```diff
@@ -415,6 +432,42 @@ The [`char`][char] type is UTF-32. If you are sure you need to do a codepoint-at
...
+This can be done with the [`Into`][into] trait, as in the following example, which lets code be generic over conversions. Note that you will need to call the `.into()` method within the function the bound is placed on.
+
+```rust
+fn accepts_both<S: Into<String>>(s: S) {
+    let s = s.into();   // This will convert s into a `String`,
+                        // regardless of the input type.
+    // ... the rest of the function
+}
+```
```
brson: This is not the answer I expected! I'd say `AsRef<str>` is the more common way to do this.

faq.md
AndrewBrinker: Yeah, I wasn't quite sure how to handle those answers. The linked article is mine, and I figured just linking to it would be easiest, but I wasn't sure if an answer like that would be alright.

faq.md
```diff
@@ -498,7 +551,7 @@ fn main() {
 What is the difference between passing by value, consuming, moving, and transferring ownership?
 </a></h3>
 
-These are different terms for the same thing. In all cases, it means the value has been moved to another owner, and moved out of the possession of the original owner, who can no longer use it.
+These are different terms for the same thing. In all cases, it means the value has been moved to another owner, and moved out of the possession of the original owner, who can no longer use it. If a type implements the `Copy` trait, the value will be copied rather than moved, and so the original owner can continue to use the value after the function call.
```
brson: "after the function call". The surrounding text doesn't mention function calls. Maybe this can just end with "can continue to use the value", or "can continue to use the original variable".

faq.md
```diff
@@ -797,6 +850,12 @@ The [`File`][File] type implements the [`Read`][Read] trait, which has a variety
 For buffered reads, use the [`BufReader`][BufReader] struct, which helps to reduce the number of system calls when reading.
 
 <h3><a href="#how-do-i-get-command-line-arguments" name="how-do-i-get-command-line-arguments">
```
brson: Wrong anchor name.

faq.md
```diff
@@ -1210,6 +1269,77 @@ Converting a C-style enum to an integer can be done with an `as` expression, lik
...
+    }
+}
+
+#[lang="eh_personality"] extern fn eh_personality() {}
+#[lang="panic_fmt"] fn panic_fmt() -> ! { loop {} }
+#[lang="stack_exhausted"] extern fn stack_exhausted() {}
+```
+
+Which should indeed roughly match C in memory usage.
+
```
brson: This explanation misses the major source of bloat in Rust, monomorphization.

brson:
cc @rust-lang/libs @rust-lang/lang new faqs that you may have opinions on.

faq.md
AndrewBrinker: Yeah, I suppose it depends on what you're looking for. There's `Into<String>` if you know you'll need an owned string, there's `AsRef<str>` if you know you'll need a slice. Then there's `Cow<str>` if you really don't care, and want to treat the value uniformly. Should the answer mention all of them?

faq.md
AndrewBrinker: I've added a note about monomorphisation (what's the spelling on that? The rest of the FAQ uses an 's', but I know 'z' is often used). Do you think the answer deserves a more substantial rewrite?

faq.md
solson: I would mention `Into<String>` and `AsRef<str>`.

faq.md
```diff
@@ -498,7 +540,7 @@ fn main() {
 What is the difference between passing by value, consuming, moving, and transferring ownership?
 </a></h3>
 
-These are different terms for the same thing. In all cases, it means the value has been moved to another owner, and moved out of the possession of the original owner, who can no longer use it.
+These are different terms for the same thing. In all cases, it means the value has been moved to another owner, and moved out of the possession of the original owner, who can no longer use it. If a type implements the `Copy` trait, the value will be copied rather than moved, and so the original owner can continue to use the value.
```
solson: "copied rather than moved" is potentially misleading since moves generally involve copying the memory (modulo optimizations).

faq.md
```diff
@@ -1210,6 +1258,78 @@ Converting a C-style enum to an integer can be done with an `as` expression, lik
...
 Converting in the other direction can be done with a `match` statement, which maps different numeric values to different potential values for the enum.
 
+<h3><a href="#why-do-rust-programs-use-more-memory-than-c" name="why-do-rust-programs-use-more-memory-than-c">
+Why do Rust programs use more memory than C?
+</a></h3>
+
+There are several potential explanations for a Rust program using more memory than a C program:
+
+- The Rust program wasn't compiled in release mode.
+- The Rust program is using jemalloc (the default allocator) rather than the system allocator.
```
solson: Is there a jemalloc-related item in this FAQ or the book we can link to explain why jemalloc is the default?

faq.md
nagisa: > You would take a String by-value and then only use a slice that you get from it.

There’s `impl AsRef<str> for str`

faq.md
nagisa: z = US English, s = every other English. As always.

faq.md
solson: @nagisa Sure, that's the case where it's fine. I'm complaining about `impl AsRef<str> for String` which would let you pass a `String` even though the callee will presumably only use the `&str` it gets from `.as_ref()`.

nagisa: `AsRef<str>` is still strictly more general than &str. You could implement it for variety of types 
which you otherwise would like to not have Deref<Target=str> implemented.

faq.md
solson: Canada uses `z`, too! :stuck_out_tongue:

(I selfishly wish everyone just standardized on `z` because I once spent half an hour trying to fix header includes because I didn't notice that I was calling `FreeImage_Initialize` when it was actually called `FreeImage_Initialise`. :disappointed:)

faq.md
solson: I suppose that's true. I guess the pattern might be `s: &S where S: AsRef<str>` - I hadn't considered that.

faq.md
AndrewBrinker: Right, my question wasn't about what the English rule is. It was about which one Rust prefers.

faq.md
```diff
@@ -415,6 +428,35 @@ The [`char`][char] type is UTF-32. If you are sure you need to do a codepoint-at
...
+Rust has four pairs of string types, [each serving a distinct purpose](http://www.suspectsemantics.com/blog/2016/03/27/string-types-in-rust/). In each pair, there is an "owned" string type, and a "slice" string type. The organization looks like this:
+
+|               | "Slice" type | "Owned" type |
+|:--------------|:-------------|:-------------|
+| UTF-8         | `str`        | `String`     |
+| C-compatible  | `CStr`       | `CString`    |
+| OS-compatible | `OsStr`      | `OsString`   |
+| System path   | `Path`       | `PathBuf`    |
+
+Rust's difference string types serve different purposes. `String` and `str` are UTF-8 encoded general-purpose strings. `OsString` and `OsStr` are encoded according to the current platform, and are used when interacting with the operating system. `CString` and `CStr` are the Rust equivalent of strings in C, and are used in FFI code, and `PathBuf` and `Path` are convenience wrappers around `OsString` and `OsStr`, providing methods specific to path manipulation.
```
killercup: > Rust's difference string types serve different purposes.

s/difference/different or maybe 'various' so there is only one 'different' in that sentence

faq.md
AndrewBrinker: Yeah, probably better to just say that the original owner isn't invalidated.

faq.md
AndrewBrinker: There isn't, but there should be. I'm not familiar with the reasoning behind the jemalloc choice, so I'll open an issue to add an answer about it.

AndrewBrinker: Alright, I've opened #395.

faq.md
AndrewBrinker: Yeah, I'll fix the typo, but I actually like the repetition in the sentence. If other people feel strongly opposed, I'll change it.

AndrewBrinker:
Alright, I've gone through the latest round of feedback. Let me know if anything else needs fixing!

brson:
Thanks @AndrewBrinker.

PR #397, opened 2016-06-04
FAQ: prefer to use "str".to_string()

liigo:
to_string() is more straitforward than to_owned(), IMO. and it is not slower anymore after Rust 1.9 stable.


steveklabnik:
Yup, this is considered official style. Thanks!

PR #403, opened 2016-06-08
Reference C++ templates in section on monomorphization

djc:


AndrewBrinker:
The policy has been not to compare Rust to other languages outside of the "Other Languages" section. People reading the FAQ come from lots of different backgrounds, and we can't assume they know C++. My preference is to continue with that policy.

djc:
This is actually part of the "Other languages" section, though?

AndrewBrinker:
Oh geez, you're right. This is why I shouldn't try to give feedback on PRs right after waking up.

It's probably fine to reference C++ here, although I worry that people will make incorrect assumptions about how Rust traits and C++ templates compare. It's probably worthwhile to add an answer explaining the differences between them.

PR #404, opened 2016-06-08
No need for title case in paragraph titles

djc:


faq.md
```diff
@@ -1321,7 +1321,7 @@ __Link-Time Optimization__
```
brson: What about the one above, "Link-Time Optimization"?

djc: I decided against it because it ends up being a somewhat well-known acronym, LTO. However, it probably makes sense to similarly lowercase it here.

brson: I think lowercase too. r=me with that change.

djc:
Fixed!

brson:
Thanks!

PR #406, opened 2016-06-08
faq: -Z not -C unstable-options

durka:
Not sure if we should add a note that _all_ `-Z` options are supposed to be unstable, though this is not yet enforced.


brson:
Thanks!

PR #410, opened 2016-06-10
Redirect VC++ Build Tools link to landing page

AndrewPardoe:
We (finally!) have a landing page for the VC++ Build Tools. I'd like to redirect the link on downloads.html from my blog post to the landing page. Also, I'd like to remove the direct download link because it's 1) easy to find on the landing page and 2) can go stale as we update the Build Tools. Note that the next option--install Visual Studio Community--has no direct download link, so removing this adds consistency.


alexcrichton:
Yay!

Issue #247, opened 2016-01-14
Modify downloads.md to use the default template


brson:
After the [next update](https://github.com/rust-lang/rust-www/pull/244) most every page on the website has a consist style with a horizontal rule and page title following the header. Except the download page, which has a very customized style. It looks weird to me when I'm clicking between pages.

fredericmarx:
I built a possible fix for this at https://github.com/fredericmarx/rust-www/commit/7ca5d95acc1c9754964a8a755a91f19cae540372.

Feel free to comment.

<img width="981" alt="screen shot 2016-03-30 at 02 37 55" src="https://cloud.githubusercontent.com/assets/954294/14128850/ef303648-f624-11e5-9328-e7da66d017b6.png">

brson:
@fredericmarx Sorry I missed your comments here. The screenshot looks good to me, except that for some reason the "Downloads" header feels out of place to me. I'm not sure if it's just because I'm so used to the page without the header or not. It seems like doesn't fit into the page as well as the others, where on the other pages the header is immediately followed be some sort of summary text.

Perhaps I would get used to it. @steveklabnik what do you think of that screen?

steveklabnik:
I feel _exactly_ the same. I'm tempted to merge it though, as I don't really dislike it, I'm just not sure if I like it, which is a subtle difference.

fredericmarx:
Ha, no worries, I also only just realised you had replied.

Any news on this? I figured as most pages had a header it would be a good idea to have one here, too, for consistency's sake. I don't have a strong opinion against leaving it out though either.

steveklabnik:
I think I vote merge. @brson ?

brson:
Yes let's do it @fredericmarx

PR #417, opened 2016-06-22
Modify downloads.md to use the default template

fredericmarx:
Solves #247 

- Modify `downloads.md` to use the default template
- Add header to `downloads.md` for additional clarity / consistency
- Use explicit `<hr>`-tags to denote section breaks
  - Level two headings had a top border by default. In some places this wasn't helpful. Using `<hr>`-tags instead solves that styling issue and also provides a more semantic section delimiter

### DIFF
--- a/community.md
+++ b/community.md
@@ -24,6 +24,8 @@ care about making the community a safe space for you.
 [mod_team]: https://www.rust-lang.org/team.html#Moderation
 [mod_team_email]: mailto:rust-mods@rust-lang.org
 
+---
+
 ## Getting Started
 
 The most important community resources for those new to Rust are:
@@ -36,6 +38,8 @@ The most important community resources for those new to Rust are:
 You may also find help on the question and answer site, [Stack Overflow][stack_overflow].
 
 [stack_overflow]: https://stackoverflow.com/questions/tagged/rust
+
+---
 
 ## News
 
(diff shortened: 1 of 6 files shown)


AndrewBrinker:
Hmm... I don't know that I'm wild about the addition of horizontal rules everywhere. It's inconsistent with [the purpose of `<hr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr):

> The HTML `<hr>` element represents a thematic break between paragraph-level elements (for example, a change of scene in a story, or a shift of topic with a section). In previous versions of HTML, it represented a horizontal rule. It may still be displayed as a horizontal rule in visual browsers, but is now defined in semantic terms, rather than presentational terms.

It's not being used to separate paragraph-level content here, so I would consider the change incorrect, per [the standard](https://html.spec.whatwg.org/multipage/grouping-content.html#the-hr-element).

The changes to `downloads.md` seem fine.

brson:
I'm also unconvinced about the hr's (it would help to see this deployed to your own branch). Can we limit this PR to just updating the downloads page and consider the hr's separately?

fredericmarx:
Hi everyone, thanks for commenting!

I used the existing `.side-header h2`-rule to remove top borders for headlines on the download page so the explicit `hr`s aren't needed anymore.

brson:
Huh, I just noticed that this doesn't actually change downloads.md to use the default template. It's still using 'basic'. Is that the intent @fredericmarx? I hoped we would end up on the same template as the other pages.

fredericmarx:
@brson certainly wasn't the intent. Don't know what happened there; fixed now.

brson:
Thanks @fredericmarx!

PR #424, opened 2016-06-30
Add timeout to playpen requests

skade:
This adds a timeout of 2 seconds to all playpen requests. If playpen doesn't respond, assume it to be down.

Error message could be better phrased, maybe.

### DIFF
--- a/_includes/editor.js
+++ b/_includes/editor.js
@@ -92,6 +92,8 @@
       code: program
     });
 
+    req.timeout = 6000;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Right now the timeout on the playpen itself is 5 seconds, perhaps this could be bumped to maybe 6-8 seconds?

skade:
Fixed. Upped it to 6.

PR #431, opened 2016-07-07
Detect 32 bit windows

### LINKED ISSUE #303: 32-bit Windows platform detection
Moved from https://github.com/rust-lang/rust/issues/20087

https://github.com/rust-lang/rust-www/blob/master/_includes/set_platform.js does not seem to take into consideration 32-bit Windows, which isn't as nice as it could be.

matklad:
Based on http://stackoverflow.com/questions/1741933/detect-64-bit-or-32-bit-windows-from-user-agent-or-javascript
and http://www.useragentstring.com/pages/useragentstring.php

Closes #303 

Unfortunately I was unable to test this locally, `jekyll serve` failed on both windows and linux, with different errors :( Is there some sort of staging site maybe?

### DIFF
--- a/_includes/set_platform.js
+++ b/_includes/set_platform.js
@@ -6,7 +6,13 @@
     if (navigator.platform === "Linux x86_64") { return "x86_64-unknown-linux-gnu";}
     if (navigator.appVersion.indexOf("Linux")  !== -1) { return "x86_64-unknown-linux-gnu";}
     if (navigator.appVersion.indexOf("Mac") !== -1) { return "x86_64-apple-darwin";}
-    if (navigator.appVersion.indexOf("Win") !== -1) { return"x86_64-pc-windows-gnu";}
+    if (navigator.appVersion.indexOf("Win") !== -1) {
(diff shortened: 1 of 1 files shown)


steveklabnik:
> Unfortunately I was unable to test this locally, jekyll serve failed on both windows and linux, with different errors :( Is there some sort of staging site maybe? 

There is not. What errors did you get?

matklad:
Ok, I was able to test this on 32 bit windows in Chrome and IE and it worked :) I don't currently have access to 64 bit windows though. 

The error on windows was about missing `redcarpet` dependency. I was able to fix it by installing something called `DevKit` and adding `gem 'redcarpet'` line to `Gemfile`.

matklad:
Linux x64 did not regressed as well. I got the error initially because I had a jekyll installed by the ubuntu package manager. Removing it helped.

alexcrichton:
Awesome, thanks! @brson wanna merge this tomorrow or do you think we should merge now (30m before the release)

brson:
now is fine

alexcrichton:
woo!

PR #454, opened 2016-07-20
Use absolute URL for menu items

limeburst:
Menu items won't link to their correct pages in some cases.

Example: https://www.rust-lang.org/404/

### DIFF
--- a/_layouts/basic.html
+++ b/_layouts/basic.html
@@ -26,10 +26,10 @@
           </a>
 	</li>
 	<li class="col-xs-12 col-md-10 menu">
-	  <h2><a href="documentation.html">Documentation</a></h2>
-	  <h2><a href="community.html">Community</a></h2>
-	  <h2><a href="downloads.html">Downloads</a></h2>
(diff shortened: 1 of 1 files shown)


steveklabnik:
Hm, I would prefer not to do this, as it means that every user using the main menu needs a redirect.

@edunham @alexcrichton how is the 404 page handled? I don't even see it in this repo.

frewsxcv:
> how is the 404 page handled?

https://github.com/rust-lang/rust-www/blob/master/error.html

steveklabnik:
Ah interesting. I wonder why it's coming up as `/404` in the URL though.

limeburst:
@steveklabnik Ah, that was just me giving an example of a nonexistent page with a trailing slash.

steveklabnik:
Ohhhhhhhh I see now. Hm.

alexcrichton:
Yeah I'm not sure what the best solution would be for this otherwise, we can perhaps refine the logic over time but I'm gonna merge for now to at least prevent the breakage.

PR #461, opened 2016-07-28
Add jseyfried to compiler team

nikomatsakis:


brson:
I also added them to the irlo compiler group.

PR #463, opened 2016-07-29
Introduce alumni 'team' and move Gankro and huonw to it

brson:
This modifies the teams page to add a final 'pseudo-team' called "Rust team alumni". Their responsibility is "enjoying a leisurely retirement". This adds another piece of metadata only used by alumni: 'ex-teams', and adds it under the 'irc' hover. For Huon e.g. it says 'teams: core, lang, libs'.

We might come up with a more badass 'responsibility'. After this merges I'll remove them from the GitHub teams, mail ailiases, and irlo groups, then eulogize them on the irlo.

r? @steveklabnik 

cc @rust-lang/compiler @huonw @Gankro

### DIFF
--- a/en-US/team.md
+++ b/en-US/team.md
@@ -34,9 +34,11 @@ people:
     name: Erick Tryzelaar
   Gankro:
     name: Alexis Beingessner
+    ex-teams: ["libs"]
   huonw:
     name: Huon Wilson
     irc: huon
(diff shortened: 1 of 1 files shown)


brson:
er cc @rust-lang/core, not compiler.

steveklabnik:
I like it.

PR #465, opened 2016-08-03
Fix broken link to tutorial in FAQ

nickrolfe:
I noticed the FAQ was linking to the "Learn Rust" section of the book, but it looks like the book got rewritten a while back, and now you just get a 404 error.

I've tried to fix the link here with an appropriate replacement.


alexcrichton:
Thanks!

PR #469, opened 2016-08-07
Provide link in case redirect fails

niconii:
Browsers using uMatrix or uBlock to block JavaScript appear to block `<noscript>` tags as well.

https://github.com/gorhill/uMatrix/issues/319
https://github.com/gorhill/uBlock/issues/308

In other words, neither `<script>` nor `<noscript>` can be relied on to work. This PR adds a body to the redirect page which provides a link, in case the redirect fails.


alexcrichton:
Thanks!

PR #480, opened 2016-08-14
Add missing operators that can be overloaded

djui:
Following the implementation of [RFC 953](https://github.com/rust-lang/rfcs/blob/master/text/0953-op-assign.md) these traits were missing and should be added for completeness.


brson:
Thanks!

PR #483, opened 2016-08-18
Recursively invalidate all files and dirs

brson:
This fixes a bug where /en-US/ was not invalidated.

r? @steveklabnik

### DIFF
--- a/travis/inval-list.py
+++ b/travis/inval-list.py
@@ -16,18 +16,24 @@ def pretty_time():
     return '-'.join(time.strftime('%c').split())
 
 def list_files(site, name):
-    base = os.path.join(site, name)
-    return [os.path.join(name, f) for f in os.listdir(base) if os.path.isfile(os.path.join(base, f))]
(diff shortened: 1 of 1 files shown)


steveklabnik:
:+1: :heart:

travis/inval-list.py
```diff
@@ -16,18 +16,24 @@ def pretty_time():
...
+    for f in os.listdir(base):
+        path = os.path.join(base, f)
+        if os.path.isfile(path):
+            files.extend([name + f])
+        if os.path.isdir(path):
+            files.extend(list_files(site, name + f + "/"))
 
-def build_payload(site):
-    files = [ f for f in os.listdir(site) if os.path.isfile(os.path.join(site,f)) ]
+    return files
```
frewsxcv: Not sure if it would have helped here, but [`os.walk`](https://docs.python.org/2/library/os.html#os.walk) is a thing.

Issue #459, opened 2016-07-26
Emails on Teams page is invalid


Havvy:
The email links on the teams.md page show eg. `mailto:rust-mods [at] rust-lang [dot] org` and are invalid.

I tried to fix this, but I'm not sure what's adding the link. There's no actual link in the team.md for the link.

steveklabnik:
I believe this is an intentional anti-spam measure. I'd be in favor of just changing it to an email. 

It's in the YAML block at the top of https://raw.githubusercontent.com/rust-lang/rust-www/8c30aeb0be86dcc8b405f79f737a008209b05cb5/en-US/team.md

brson:
Fixed.

PR #485, opened 2016-08-19
Use actual emails

Havvy:
As per #459 I think @steveklabnik means that these email addresses should just be real email addresses.


alexcrichton:
Seems like this'd be up to the community team and @rust-lang/moderation teams?

erickt:
I'm in favor of switching out the mailto links with proper emails. I'm pretty sure spambots are clever enough to parse `rust-mods [at] rust-lang [dot] org`, so we're just inconveniencing people who want to contact us for legitimate reasons.

Manishearth:
wfm. google groups' spam filters are pretty good, I only recall one false positive so far.

brson:
Fine by me.

Issue #477, opened 2016-08-13
Add rotating friends section to front page


brson:
We have a lot of [friends](https://www.rust-lang.org/en-US/friends.html) now. When we set up the page we promised to promote it harder once it got going. Good for us to show Rust is used for real stuff, good for their visibility as well. By being discoverable on the front page it will encourage more friends.

Add a section to the front page that displays 3 random friends, just like the friends page does now, along with some text indicating these are "Friends of Rust", and linking to the friends page.

Fitting this into the front page in a clean way will take some love and care.

pragyes31:
Hey Brson, I would like to work on this issue.

maciejhirsz:
I could put something together there.

brson:
@maciejhirsz You got it!

PR #486, opened 2016-08-21
Add random 3 ‘Friends of Rust’ to the front page

maciejhirsz:
#477 Can spend some more time tweaking the styling, though it seems to be quite alright reusing the same CSS.

r? @brson

### DIFF
--- a/_includes/friends_index.html
+++ b/_includes/friends_index.html
@@ -1,6 +1,6 @@
 
 <div id="users" class="row"></div>
-<h2 class="friends-more">See more <a href="/friends.html">uses of Rust in production</a></h2>
+<h2 class="friends-more">See more <a href="/friends.html">uses of Rust in production</a>.</h2>
 
 <script type="text/javascript">
 
@@ -8,11 +8,7 @@
 'use strict';
 
 var users = [
-{% for user in site.data.users %}{
-  url: "{{user.url}}",
-  logo: "../user-logos/{{user.logo}}",
-  name: "{{user.name}}"
-},
+{% for user in site.data.users %}["{{user.url}}", "{{user.logo}}", "{{user.name}}"],
 {% endfor %}];
 
 function getRandomThree() {
@@ -24,19 +20,21 @@ function getRandomThree() {
 }
 
 function renderUser(user) {
-  var logo = document.createElement('div');
+  var el = document.createElement('div');
 
-  logo.innerHTML = '\
+  var url = user[0], logo = user[1], name = user[2];
+
+  el.innerHTML = '\
     <div class="user-container">\
-      <a href="' + user.url + '" rel="nofollow" alt="' + user.name + '">\
-        <img src="' + user.logo + '">\
+      <a href="' + url + '" rel="nofollow" alt="' + name + '">\
+        <img src="../user-logos/' + logo + '">\
       </a>\
     </div>\
   ';
 
-  logo.className = 'col-md-4';
+  el.className = 'col-md-4';
 
-  return logo;
+  return el;
 }
 
 var frag = document.createDocumentFragment();
--- a/_includes/friends_index.html
+++ b/_includes/friends_index.html
@@ -1,0 +1,84 @@
+<div class="site-header-with-subhead">
+  <h2><a href="/friends.html">Friends of Rust</a></h2>
+  <p>(Organizations running Rust in production)</p>
+</div>
+
+<div id="users"></div>
+
+<div class="more-friends"><a href="/friends.html">All Friends of Rust</a></div>
+
+<script type="text/javascript">
+
+(function() {
+'use strict';
+
+var users = [
+{% for user in site.data.users %}{
+  url: "{{user.url}}",
+  logo: "../user-logos/{{user.logo}}",
+  name: "{{user.name}}",
+  how: "{{user.how}}"
+},
+{% endfor %}];
+
+function getRandomThree() {
+  users.sort(function(a, b) {
+      return Math.random() - 0.5;
+  });
+
+  return users.slice(0, 3);
+}
+
+function renderUser(user) {
+  var logo = document.createElement('div');
+
+  logo.innerHTML = '\
+    <div class="user-container">\
+      <a href="' + user.url + '" rel="nofollow" alt="' + user.name + '">\
+        <img src="' + user.logo + '">\
+      </a>\
+    </div>\
+    <div class="details user-detail fade-out"\
+      <p><em>' + user.name + '</em> : ' + user.how + '</p>\
+    </div>\
+  ';
+
+  logo.className = 'col-md-4';
+
+  logo.onmouseover = function() {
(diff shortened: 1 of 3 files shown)


steveklabnik:
Can you maybe host this somewhere so we can check it out?

maciejhirsz:
Here it is: http://rust.terhix.com/en-US/

brson:
Oh this is awesome.

It's almost perfect the way it is, but a few observations:

With all the whitespace when there is no hover the "all friends of rust" looks pretty disconnected from the rest of the page. Since the "Friends of Rust" header links to the same place, can we try just removing the "all" link?

Once we do that we've got just the one header line to express everything we can. Right now it is "Friends of Rust (organizations using Rust in production)". When I read that I think it's quite verbose for what it is conveying. But at the same time that's exactly what the 'friends' page says, so it makes sense.

I like what the [ember page says](http://emberjs.com/) for this, "who's using ember?". It gets the point across with few words, but unfortunately on our page we used different terminology, the reason being (as i recall) we wanted to emphasize the 'production-ness' of the users, vs. e.g. open source organizations.

I'm inclined to try this: center "Friends of Rust" and remove "organizations running Rust in production". It won't be clear what 'friend' means, but if they click through it will be made obvious.

Do you mind making those changes? This will probably require a few rounds of iteration.

Thanks @maciejhirsz !

maciejhirsz:
Sure thing! Don't have time right now, but I'll get to it.

steveklabnik:
Oh this is so cool!

One issue with this on my screen:

2016-08-22-165657_922x339_scrot

that is, it _just_ cuts off before the "all friends" thing. Not sure if that's something that can be fixed, exactly....

nikomatsakis:
I think it could be made clearer that this is a *subset* of the friends of Rust -- I didn't find that obvious at first. I'm not sure the best way to do that though. :)

maciejhirsz:
Applied @brson's suggestions, updated version at the same url.

brson:
Thanks @maciejhirsz. Can we try one more iteration?

I have a few outstanding concerns: that 'friends of rust' doesn't indicate clearly enough what it means, nor that there are more than just three; that the centered header in that location feels unbalanced.

So I think it makes sense to say something like "See more Rust production users", but it really only makes sense to say that _after_ we've shown them _some_ production users (like your original footer). But putting that footer there makes the hover whitespace looks super wierd.

So can we try: removing the hover descriptions entirely (they can see them on the friends page); moving the "Friends of Rust" header to beneath the logos, leaving about as much vertical whitespace as the current header takes up now; changing "Friends of Rust" to "See more uses of Rust in production", with only "uses of Rust in production" being a link (so that the contrast between "see more" and "uses of..." strongly indicates it can be clicked.

Thanks for your patience.

maciejhirsz:
Updated, it does feel like an improvement. I'm tempted to add a period after the link now that it's more of a sentence.

I don't mind iterating further with this and trying things out. At times I might be slow to respond with changes :).

brson:
@maciejhirsz I agree it's begging for a period. Please add.

With that I think this is a pretty great checkpoint. Let's deploy and live with it a while and see what people think.

brson:
@maciejhirsz Oh, I'm noticing that I see Xero every time I load the page. Do you see that? Is there an error in the math?

aturon:
cc @rust-lang/core

jonathandturner:
@brson - maybe it's just your browser?

screen shot 2016-08-23 at 3 12 48 pm

maciejhirsz:
@brson Nope. The randomizer is pretty dumb - I grab the array and sort it randomly, then grab first three elements.

Could make it pick 3 random indexes instead to make it faster, but as far as the correctness goes it should be solid.

jonathandturner:
@maciejhirsz - after playing with it a little more I'm starting to see the same pattern.  Maybe we should be on the safe side and switch up the way we're picking them

maciejhirsz:
Weird, let me try picking the indexes instead...

aturon:
I like the idea of this change, but the current version gives me mixed feelings. In particular, it makes the home page feel a bit cluttered, as also feels a bit as if we're displaying advertisements. (Note also that the logos are as big, or bigger, than our own logo!) I wonder if there's some different way of styling this that would make it feel more natural?

I think part of the issue may be that the hope page tries to fit in a literal page, which has kept it very focused. This feels like a departure, and might work better in the context of a page with more sections, e.g. the [Typescript home page](https://www.typescriptlang.org/), where we could more clearly delineate. We could reap some additional benefits by having some other home page sections, but obviously that would be part of a more substantial revamp...

jonathandturner:
@aturon - to make it feel less like an advertisement, what if there was a line between the friends and the bottom of the example.  something like:

```
               --------- Friends of Rust ---------
              [ image]          [ image]          [ image]
````

maciejhirsz:
I feel less certain about [myself trying to be clever here](https://github.com/maciejhirsz/rust-www/blob/cc2704ff905acd958977e2161173768b89d33887/_includes/friends_index.html#L14-L28) than I did with the simple shuffle + slice, let me know if it seems wonky.

Re the design: a separator alike one present on other pages above headlines might help. Even without a headline it should make the 3 logos and the text below them seem more like a single unit unrelated to the rest of the page.

brson:
OK, sorry for the delay.

The core team had some high-bandwith discussion about this and unfortunately want to scale back our ambitions for now.

The problem we see is that with the current minimal front-page design, putting these logos on their gives them _very strong_ prominence. They are bigger than the Rust logo! If you compare our page to e.g. [typescript](http://typescriptlang.org/), they are way less in-your-face. And the reason for this seems to be pretty fundamental - they've just got a lot more content, so the page is structured into sections, where the later sections are clearly less important. On the Rust page, you look at it, and the impression is that these companies are of _prime importance_ to Rust.

So until we are able to do a wholistic redesign, the desire is to not use logos.

As an alternative, what if we figure out how to adjust the 'sales pitch' section, and change the text of 'show me!' to something else that links to the friends page?

Some options:

- "See how Rust is used in production."
- "See who's using Rust in production."
- "See who's using Rust."
- "And it's used for serious business, in the real world."

The style and layout may need to be tweaked to make it pretty. Not sure.

cc @rust-lang/core any opinions?

alexcrichton:
I personally like "See who's using Rust", short and sweet and gets the point across

aturon:
> I personally like "See who's using Rust", short and sweet and gets the point across

+1 on that!

To restate my earlier point, it's not just about the prominence of the logos, but the fact that the overall web design just doesn't lend itself well to this kind of addition (without it strongly feeling like a cluttered ad). I think we definitely want something like this original PR eventually, but as part of a larger revamp. Doing it textually for now is a good incremental step.

brson:
Closing in favor of https://github.com/rust-lang/rust-www/pull/536. Sorry for the hassle @maciejhirsz. It's entirely my fault.

PR #491, opened 2016-08-25
Begin Russian translation

mkpankov:
This is first spike of Russian translation.

I added a layout for non-translated pages, which is a stub saying "This page is not translated. Please help us translate <GitHub .md file edit link> or read English version <link>".

The hope is to facilitate translation by making non-translated pages visible. I also got rid of the verbatim copies of English pages in the text of untranslated pages so that there's no duplicated stuff that needs to be updated in sync with original.


steveklabnik:
>  Should I just leave off non-translated md's?

This is a good question. I'm not sure what the right answer is.

mkpankov:
@steveklabnik I checked what happens when `ru-RU/documentation.md` isn't present - it causes 404. So I left the untranslated pages and translated the `index.html`.

For future work, maybe it makes sense to default to `en-US` in case there's no file present in the locale? So that it would redirect.

steveklabnik:
Yeah, I think that makes sense.

mkpankov:
Updated the PR. I'm gonna put all the details into edited description and make it no longer WIP. Think it's ready.

steveklabnik:
Great! Thanks so much.

PR #497, opened 2016-08-30
add snips to friends of rust

kali:
Hi !

We would be glad to appear on the list.

Please use this address as point of contact: mathieu.poumeyrol@snips.ai .

Thanks.


steveklabnik:
Wonderful! Thanks so much.

PR #502, opened 2016-09-01
Link to this file on github

carols10cents:
These links went to the file that is now just a redirect, which is
confusing when you expect to see the list of meetups.

I'm not sure how to make this work in a nicer way with the translations,
but there's only two right now...


steveklabnik:
I agree that this isn't the best, but is the kind of thing that should shake out of more improvements here. Thanks!

PR #503, opened 2016-09-01
Add pittsburgh

carols10cents:
I've had a low-key meetup going on for a while, but I recently put it up on meetup.com so now I have something to link to!

Also, I organized the USA meetups alphabetically by state so it's easier to find meetups nearby :)


steveklabnik:
merged! :heart:

PR #508, opened 2016-09-04
Allow the release dates to be translated

lifthrasiir:
Use the ISO 8601 format in the config (probably not necessary as Ruby `Time.parse` is lax, but allows graceful degradation when not localized) and put some Liquid hacks [1] to properly display the localized date. Also removed an annoying space between date and comma ;)

Note: I've looked the Portuguese date format up from [ICU](http://www.localeplanet.com/icu/pt-BR/), but I don't know if I'm correct. Anyone care to review?

[1] Inspired from http://alanwsmith.com/jekyll-liquid-date-formatting-examples (but this one is better).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,7 @@
 name: The Rust Programming Language
 stable: "1.11.0"
-stable_date: "August 18, 2016"
+stable_date: "2016-08-18"
 stable_blog: "/2016/08/18/Rust-1.11.html"
 beta: "1.12"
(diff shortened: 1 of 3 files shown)


steveklabnik:
@lazpeng could you comment on this?

lazpeng:
It's correct. Also, it's really nice having the date translated to a culture-friendly format, thanks.

steveklabnik:
Great! Thanks both.

PR #509, opened 2016-09-04
Allow the localized messages to go along with the HTML

lifthrasiir:
Previously dynamic messages were hard-coded in the script. As the HTML is already localized (or we at least assume so), it is better to put them into the HTML itself---in this case with HTML5 `data-*` attributes.

This change mainly concerns with playpen messages (currently "Running..." only) and the current version replacements (which didn't really work for untranslated languages anyway---check the browser console). Other kinds of dynamic messages or omitted ones (not sure how to deal with the playpen errors in general, seems not working) can be added in the same way.

### DIFF
--- a/_includes/editor.js
+++ b/_includes/editor.js
@@ -23,6 +23,9 @@
   var errorColor = "#F6E2E2";
   var warningColor = "#FFFBCB";
 
+  // Message to show when the program is running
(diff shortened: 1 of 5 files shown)


steveklabnik:
I like it, thanks!

Eventually, it'd be good to pull this out into some kind of config file... someday.

PR #506, opened 2016-09-02
adding a pre tag to use monospace font

rjgoldsborough:
fixes #490 

adds a simple pre tag and uses it

PR #510, opened 2016-09-05
Fix editor's parsing of new error format.

### LINKED ISSUE #490: Errors displayed after running front-page code look bad
They are not in an equal width font and have no colors.

SergioBenitez:
Closes #490. Here's what it an error looks like with this PR:

screen shot 2016-09-04 at 8 47 15 pm

### DIFF
--- a/_includes/editor.js
+++ b/_includes/editor.js
@@ -183,28 +183,38 @@
     // Getting list of ranges with problems
     var lines = message.split(newLineRegex);
 
-    // Cleaning up the message: keeps only relevant problem output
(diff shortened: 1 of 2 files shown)


brson:
Huh, this is quite a different solution than the one proposed in https://github.com/rust-lang/rust-www/pull/506. It looks like this one fixes some additional problems than just the missing monospace font, but also fixes some highlighting issues. Does that sound right?

SergioBenitez:
That's right. It also doesn't mess with the HTML add at all.

PR #521, opened 2016-09-15
Remove client00 from Mibbit URLs

Havvy:
By not going to `client00` directly, users go through a load balancer that can put them on any available client server. It's not always guaranteed at `client00` will be available.

(I'm an oper on the Mibbit network, I know these things.)


brson:
Thanks @Havvy !

PR #523, opened 2016-09-15
Add the style team

nrc:


nrc:
It seems to me to be pure win - we advertise the existence of the team and make the team feel more official. I don't think it takes anything away from the more permanent teams, there is no maintenance overhead, and there is no cost to having more data on this page.

brson:
I've gone back to both the [style strike team RFC](https://github.com/rust-lang/rfcs/pull/1607) and the [unsafe strike team RFC](https://github.com/rust-lang/rfcs/pull/1643) to see if there was a definitive word on whether we are adding strike teams to the web page, and there is not, though both [myself](https://github.com/rust-lang/rfcs/pull/1607#issuecomment-224734752) and [niko](https://github.com/rust-lang/rfcs/pull/1607#issuecomment-224735807) suggested we did not expect strike teams to be listed on the web site.

I am starting to come around to the idea that we should be more expansive in how we define teams, but I'm frustrated about how we've gotten here, that we did not sufficiently establish a definition that we all agree on before creating them.

nrc:
> that we did not sufficiently establish a definition that we all agree on before creating them.

Yeah, I guess this is the fundamental problem - I think we all kind of have slightly different ideas of what the strike teams should be like and it is probably worth discussing this somewhere since I expect we'll create more in the future.

My personal mental model has been that strike teams are just like regular teams, but have limited duration and a very specific focus. It seems you and Niko are thinking of something more lightweight from the linked comments.

In practical terms, I think that if we don't reuse the existing procedures and communication channels for teams, then we have to create new ones for strike teams, and that is a lot more effort than the administration of adding and later removing things like this. It is feeling difficult to get the style/formatting process the visibility it needs, and I feel very out of touch with what the unsafe guidelines team are doing.

aturon:
@brson @nikomatsakis Are there specific downsides you see with listing strike teams on the web site?

Personally, I'm in favor of listing them, and think we should also highlight them on the eventual roadmap page. That's both to recognize the teams in an official way, and to help direct people toward important work that's under way.

nikomatsakis:
I'm ok with it; maybe we can link the team name over to some description of its purpose (or even the RFC). I just didn't expect to do it, but I'm not entirely sure why. It seems to make sense now. =)

brson:
Merged.

PR #526, opened 2016-09-16
Remove the BOM

mkpankov:
Whoops we did it again :)

@E100Beta please make sure there are no Unicode Byte Order Marks in Markdown files in further PRs. Jekyll doesn't process such files correctly :(

### DIFF
--- a/ru-RU/documentation.md
+++ b/ru-RU/documentation.md
@@ -1,4 +1,4 @@
-﻿---
+---
 layout: ru-RU/default
 title: Документация Rust &middot; Язык Программирования Rust
 ---


mkpankov:
BTW, I found the [issue in Jekyll](https://github.com/jekyll/jekyll/issues/2853). It was fixed in version 3.1.1 and now Jekyll should strip BOM automatically.

What version is used for rust-lang.org? Maybe worth updating?

steveklabnik:
3.1.2: https://github.com/rust-lang/rust-www/blob/master/Gemfile.lock#L6

So that's confusing.... anyway, thanks!

PR #537, opened 2016-09-24
Added Rust User Group from Leipzig

saschagrunert:
Hey, we are really new but I think the Leipzig User Group is it worth to be on the web site. :)


steveklabnik:
Thanks!

PR #540, opened 2016-09-26
添加了简体中文的连接连接

3442853561:
adding Chinese(Simplified) link

### DIFF
--- a/_layouts/basic.html
+++ b/_layouts/basic.html
@@ -40,7 +40,7 @@
       <p>Our site in other languages:
         <a href="/pt-BR/">Português</a>,
         <a href="/ru-RU/">Русский</a>,
-	<a href="/zh-CN/">简体中文</a>,
+	<a href="/zh-CN/">简体中文</a>
       </p>
     </footer>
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
r? @steveklabnik 

Although looks like some other work may need to be done on the infrastructure? Right now the site doesn't seem to render well :(

https://www.rust-lang.org/zh-CN/

GuillaumeGomez:
There is a syntax error (an unexpected '%').

steveklabnik:
> Although looks like some other work may need to be done on the infrastructure? Right now the site doesn't seem to render well :(

Yikes!

steveklabnik:
It seems that jekyll is just straight-up copying `zh-CN/index.html` into the `_site` directory rather than doing the templating. I have no idea why.

steveklabnik:
Okay, it has to be something specific, becuase if I copy the english one over into the zh-CN folder, it does render it. wtf.

steveklabnik:
Okay, I fixed it: https://github.com/rust-lang/rust-www/pull/541

Now that it's rendering properly, I think this is a good idea. Thanks!

PR #559, opened 2016-09-30
Add a language footer to the French basic template

benoit-l:
Allows visitors to switch back to English (or any other language) if they get the French version of the website.


alexcrichton:
Thanks!

PR #574, opened 2016-10-12
add withoutboats to the language-design team

nikomatsakis:
@withoutboats - is the name "Without Boats" what you would prefer? That's what I used. :)


nikomatsakis:
Err, my question wound up in the commit message. Not quite what I wanted. Bah. Let me try again.

PR #575, opened 2016-10-12
add withoutboats to the language-design team

nikomatsakis:
@withoutboats - is the name "Without Boats" what you would prefer? That's what I used. :)


withoutboats:
Works for me :-)

PR #578, opened 2016-10-14
There are something new in Chinese translation.

3442853561:
一些中文翻译的新玩意儿


zh-CN/downloads.html
```diff
@@ -9,12 +9,11 @@
     <div class="row install">
       <div class="col-md-4 side-header">
         <h2 id="stable">{{ site.stable }}&nbsp;</h2>
-        <h3>{{ site.stable_date | date: "%B %-d, %Y" }}</h3>
+        <h3>{{ site.stable_date | date: "%Y 年 %m 月 %-d 日" }}</h3>
```
lifthrasiir: Shouldn't this be `%Y年%-m月%-d日`, like "2017年1月1日" or "2017年1月15日"? That is, no leading zero, and no space inbetween. Also, `site.beta_date` remains untranslated.

(Tip: If you want weekdays as well, you can make use of [arrays](https://github.com/rust-lang/rust-www/blob/74af2f75c9dc9439332d9a34fa7702570a86cc92/pt-BR/downloads.html#L9-L11) as previously used by `pt-BR`. Liquid's support of array is at best strange, though.)

zh-CN/team.md
```diff
@@ -213,7 +212,7 @@ the [Rust security disclosure process](security.html).
     {% assign sitename = "github" %}
   {% endif %}
   {% assign site = page.sites[sitename] %}
-  <li class="person {% if team.lead and team.lead == nick %}lead{% endif %}">
+  <li class="person {% if team.lead and team.lead == nick %}负责人{% endif %}">
```
lifthrasiir: This is a CSS class and should not be translated.

zh-CN/downloads.html
3442853561: In Chinese there need a space between （number or letter ）and （chinese words）

zh-CN/team.md
3442853561: It has been modified

zh-CN/downloads.html
3442853561: I don't need weekdays as well,%m mean month,
and I don't know want mean about `site.beta_date`
@lifthrasiir

zh-CN/team.md
3442853561: @lifthrasiir fixed

zh-CN/downloads.html
lifthrasiir: @3442853561 My impression from various Chinese websites was that there are no spaces between numbers and Chinese letters, but if a space is needed, okay with that.

`%m` will always display two digits for month. `%Y 年 %m 月 %-d 日` will give you "2017 年 01 月 1 日" for example, which looks strange.

On `site.beta_date`, there are two dates for the latest stable release and the upcoming stable release (which is currently at beta). You will want to translate both.

3442853561: Thanks！I will fix it.

3442853561: @lifthrasiir THX,I fixed it.

alexcrichton:
Thanks!

PR #579, opened 2016-10-14
Added some symbolic languages

3442853561:
Added some symbolic languages


3442853561:
@steveklabnik Localization of the page more and more. So how about this proposal?

steveklabnik:
I am not sure that the globe is enough to recognize that this is what this is supposed to do.

brson:
I can't see this patch in context, but do suspect the colorful globe would look out of place on our spartan website.

PR #580, opened 2016-10-16
Fix typography and wording in French code example

Calinou:
- "éditable" -> "modifiable" (more correct wording)
- In French, colons should have a space before as well


steveklabnik:
Thanks! I never knew that about colons, interesting!

PR #582, opened 2016-10-16
Change "Téléchargements" for "Télécharger" in French navigation bar

Calinou:
This makes it fit on a single line on the current layout (instead of two with "Téléchargements").


alexcrichton:
Thanks!

PR #583, opened 2016-10-16
French: Change "32-bit"/"64-bit" to "32 bits"/"64 bits" on downloads page

Calinou:
Also changes "GNU ABI"/"MSVC ABI" to "ABI GNU"/"ABI MSVC" in the French downloads page.


steveklabnik:
Thanks!

PR #584, opened 2016-10-16
French: Fix typo in downloads page

Calinou:
Removes an extraneous "The".


TheoChevalier:
Note that this is already fixed in #561 ;)

brson:
Thanks, but already fixed in #561.

PR #586, opened 2016-10-17
French: Tweaks to ABI wording in downloads page

Calinou:
"GNU ABI"/"MSVC ABI" -> "ABI GNU"/"ABI MSVC" in more places (leftover from my other commit).

Sorry, it looks like I forgot some occurences of the old (incorrect in French) wording.


aturon:
Thanks!

PR #595, opened 2016-10-21
Updated number of contributors on https://github.com/rust-lang/rust/

atjonathan:


aturon:
Thanks for this!

While we're at it, can we correct the text to not just be talking about `rustc`? The repository also contains the standard library and documentation.

aturon:
Thanks!

PR #597, opened 2016-10-23
Refactor non-localizable data out of localized pages.

lifthrasiir:
- The GPG key in the security page should be identical across languages, so it now includes a copy of the existing key file.

  Liquid seems to require the included file to be in `_includes`, so the copy was necessary---the symlink also works but was avoided to avoid possible platform issues (not to say that the current website builds everywhere though).

- Almost all data in the team page has been moved to `_data`. Previously translated pages have been retrofitted to use a saner localization mechanism (modelled after what `ko-KR` was using).

  This should fix the absence of withoutboats except for `en-US` ;)


alexcrichton:
Awesome, thanks!

PR #598, opened 2016-10-23
Adjustments to pt-BR

leodasvacas:
Adjustments to home page and a typo in documentation.
ping @lazpeng

### DIFF
--- a/_includes/pt-BR/example.rs
+++ b/_includes/pt-BR/example.rs
@@ -1,4 +1,4 @@
-// Esse código é editável e rodável!
+// Esse código é editável e executável!
 fn main() {
     // Uma calculadora simples de inteiros:
     // `+` or `-` significa somar ou subtrair 1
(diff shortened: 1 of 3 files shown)


lazpeng:
@leodasvacas , valeu pelos ajustes, passou despercebido. Só umas coisinhas: talvez fosse melhor manter "segmentation fault" na home page (mesmo que os dois juntos), não sei o que você acha. E outra coisa: A minha tradução de runtime tá horrível, realmente, mas ainda pode ser melhor. Talvez deixar "runtime" mesmo com uma anotação no rodapé explicando o que é exatamente. Se for fazer isso, o mesmo com semânticas de move seria interessante. De qualquer forma, obrigado.

(For those who don't understand, I argued about what he could improve that I missed. Should I translate the full text?)

leodasvacas:
I think we should translate everything that can be reasonably translated. "falha de segmentação" is a well known translation, it's even got a [wikipedia page](https://pt.wikipedia.org/wiki/Falha_de_segmenta%C3%A7%C3%A3o). "Move semantics" is a very C++ specific term even in english, I don't think a reasonable translation exists. "sistema de execução" is a bit obscure but "sistema de execução java" gets many hits on google so I think it's a reasonable translation. I don't know how we could explain this in a footnote, but maybe we could put the original term in parenthesis after the translation.

lazpeng:
Agreed.

leodasvacas:
It seems we agree this can be made better but is an improvement as is.
r? @steveklabnik

brson:
Thanks!

PR #599, opened 2016-10-24
Add links to PGP signatures. Fixes #594.

### LINKED ISSUE #594: Give Links to pgp signature files on Download page
Currently the pgp signatures for files downloadable from the official download page are only obtainable by guessing the link by adding .asc to the actual files' download link

i.e guessing https://static.rust-lang.org/dist/rust-1.12.1-x86_64-pc-windows-msvc.msi.asc for https://static.rust-lang.org/dist/rust-1.12.1-x86_64-pc-windows-msvc.msi 

It would be better if these asc files are prominently linked on the download page. 

Edit: Also link the official signing key on the download page

udan11:


brson:
While I think this is useful information to expose, I'm not sure the best way to do it.

If we were to do this I think we need more than just links that say "PGP". What is a user supposed to do with these files? We at least would need to indicate what their purpose is, how to use them, where the key is, how to _validate the key_ , etc.

Adding these links doubles the number of links on the page for something almost nobody is going to use.

When we promote rustup.rs to be the main installation method I anticipate the installation options requiring a lot more explanation sadly, so it may make sense to think how to integrate this stuff then. I expect that the install.html page will be the dead-simple installation instructions ala https://www.rustup.rs, and having another page for other installation methods. That page will necessarily have explanation of the differences between install methods, so it could make sense to explain "if you want to validate that your binaries are the same that the Rust project release then ...".

brson:
This is superceded by https://github.com/rust-lang/rust-www/pull/621. Thank you @udan11 !

PR #600, opened 2016-10-24
Recommend rustup.rs over rustup.sh. Fixes #478.

### LINKED ISSUE #478: Recommend rustup.rs over rustup.sh
Specifically at https://www.rust-lang.org/en-US/downloads.html, and recommend rustup.rs for Windows too, right?

We've [switched the new book to do this already](https://github.com/rust-lang/book/commit/8af84e7514be2a3aa72e3a421925cf68cb73f342) :)

If it's too soon to do this because rustup.rs isn't stable enough yet, consider this a reminder issue to make this change once it is :)

udan11:


alexcrichton:
r? @brson 

Now that the checksum failure is take care of this should in theory be ready to go, right?

brson:
Thanks, but I'm not ready to do this yet, and want to do it in a more comprehensive manner. I think this page must be rewritten when moving to rustup.rs.

brson:
(I am planning to do this any day now).

PR #602, opened 2016-10-24
Recommend MSVC ABI. Fixes #341.

### LINKED ISSUE #341: Recommendation of Windows ABI
I notice we still reference that users should use the GNU ABI is 'recommended for typical uses' - however a long chat in IRC yesterday pointed out this has been the recommendation since 1.0 and a lot of work has gone into MSVC since. 

I did have look at the outstanding MSVC issues and I'd like to propose that MSVC becomes 'recommended for typical uses' due to the more stable linker and such like.

These days openssl works just fine on the MSVC ABI and you get better backtraces and debug information (hell, you even get a PDB file - awesome).

Even if we don't change the recommendation we should make the choice clearer - from IRC the Windows users definitely have less issues with it then the GNU ABI.

udan11:


alexcrichton:
r? @brson 

IIRC you wanted to change a few other things at the same time though? I may be misremembering...

PR #604, opened 2016-10-26
Translate the entire comments to pt-BR

thebergamo:
Just update the comments to PT-BR the comment, all the comments are in Portuguese, but the "OR" condition not, today this is common use in Portuguese not only in english.

cc @lazpeng 

NOTE: Not change the first line comment because already have a PR that fix the translation


brson:
Thanks!

PR #621, opened 2016-11-19
Update website for rustup installation

brson:
This patch makes rustup the default Rust installer. The main changes are in these pages.

- [Install page](http://54.215.61.108:4000/en-US/install.html)
- [Other installers](http://54.215.61.108:4000/en-US/other-installers.html#other)
- [Front page](http://54.215.61.108:4000/en-US/)

This is still rough, but is content complete, so I'm posting it to start collecting feedback.

Some of the ideas here:

- It should be obvious how to install
- The install page is tailored to each platform, with windows-specific instructions
- Most of the install and post-install instructions should be handled in rustup itself
- Only key information needed to get a working toolchain should be surfaced on the install page
- Everything else is buried on the "other installers" page
- Other installers page should include long-missing info, like links to 2nd tier platforms and signatures

On the install page you can type "test" to cycle through platform-specific stuff.

The early feedback I've gotten felt like the info here was overwhelming. So there's probably some streamlining to do. In particular, most of the content of the install page could be cut or bumped to the other-installers page.

Some caveats to be aware of during this transition:

The two pieces of information that I felt compelled to surface on the install page - that rustup is a tool you need to know about, and that the PATH needs to be configured correctly - are unfortunate. I'd prefer if we could not explain anything. The reason you need to know about the rustup command is because it's used for uninstallation, and I consider it important that we explain how to uninstall.

Today, on Windows, rustup will detect whether MSVC is available during install, and if so, use the MSVC toolchain, otherwise use GNU. I found explaining this to be too hard, and intend to make rustup just use the MSVC toolchain period, and offer guidance on how to install MSVC or change to GNU.

In review, I'd focus on the content for now more than the code. Once it's close to ready there will be more to clean up and delete, and I'll need to figure out how to deal with the translations.

r? @steveklabnik cc @aturon


steveklabnik:
I like this overall! Some thoughts:

1. the download button on the front page is huge, which looks a bit weird to me.
2. when mentioning package managers, linux stuff + homebrew, Chocolately for Windows deserves a mention
3. when are we targeting landing this? The book needs ~12 weeks to get in line.

aturon:
@brson Thanks for this! It looks great by and large, just a few nits.

### Front page

I'd suggest changing the text of the button on the front page to say "Install", at the same font size as the version number, and drop the "Current release" (which is implied). At any rate, I think mentioning "install" or "download" somewhere on the button is important, to make clear what the *action* of the button is.

Otherwise, :+1:

### Install page

This is the page that needs the most work, I think. Here are some thoughts:

- The section headers here seem to be gray, rather than black, for some reason.

- I think the top section needs a header to orient it -- just like all of the other pages on the web site. I think the top header should just be "Install Rust"

- I would drop the "Installing Rust" section entirely. I don't think it adds a lot of information, except for the bits about uninstallation, which should be discoverable via the rustup info below it.

- I'd suggest breaking out "Other installation methods" to its own top-level section. I know it'd be short, and we don't want to encourage people in this direction per se, but right now it feels *really* buried.

- The link to the rustup docs should probably go straight to the rendered readme.

### Other installation methods

This page feels a lot better than in the earlier draft -- cohesive and clean. Just a few notes:

- Should "Other ways to install via rustup" be "Other ways to install rustup"? I can see reasons for either choice...

- I think that if we include the rustup.sh script, we need to draw a LOT of attention to the fact that this is NOT the same as rustup, and say what it is. (Personally, I think it's better to just not mention it).

- The bit about "The maturity and completeness of third-party packages is unknown" felt bad. I wonder if this is a case where it's better to just not say anything -- maybe it'd be better to add this section when we feel more confident about the maturity/quality of these packages? (It's also clear that we recommend rustup, since that's where the site directs you). Overall, I'd drop this section.

aturon:
@steveklabnik 

> when are we targeting landing this? The book needs ~12 weeks to get in line.

This should go live ASAP. It's fine for the book to lag the "official" change IMO, since rustup.rs has been the de facto way to install Rust for quite a while nowl

brson:
@steveklabnik thanks for reviewing.

> the download button on the front page is huge, which looks a bit weird to me.

I agree. I'll take another pass at it.

>     when mentioning package managers, linux stuff + homebrew, Chocolately for Windows deserves a mention

OK. I will add it.

>     when are we targeting landing this? The book needs ~12 weeks to get in line.

Oh, I was thinking of landing it as soon as possible. I hadn't thought about the stable book not being in agreement with the website. I also have not prepared any changes to the book for this, at least not yet.

steveklabnik:
I sent in a PR for the book, you're the reviewer IIRC.

I would be okay with landing this before the book, getting people on rustup ASAP is more important than consistency. Maybe we can be a bit more okay with changing the book earlier in cases like these in the future, though 😄

brson:
OK I've updated based on all suggestions.

- Front page button is smaller and less verbose.
- The release information on the install page changed layouts.
- The install page has black headers and one main header.
- I reduced the information on the install page and reduced details about windows caveats. I'll try to sink some of that info into the rustup readme.
- Gave other methods its own section header.
- Removed some sections from the other methods page.

Some outstanding concerns of mine

- I need to update rustup itself to use MSVC by default, as indicated in these instructions, instead of picking the ABI based on MSVC detection.
- I need to polish the rustup docs and install experience to account for issues raised here and info not presented here.
- rustup doesn't actually have mips/powerpc/s390x builds yet. I need to either get them merged or remove from the instructions here.
- Need to do a pass checking that things look good on mobile form factors. I suspect there are bugs.
- Need to review and merge book update https://github.com/rust-lang/rust/pull/37934
- Need to deal with i18n. Probably once we're decided on the design, I should go ahead and convert all the languages.

Remember you can type "test" to see the platform-specific instructions.

r? @aturon @jonathandturner @retep998

brson:
Oh @aturon re headers on the install page. The reason for both having gray headers and not the top-level header was that I thought they were distracting the eye from the big box of instructions. In the new revision the big "Install Rust" header still feels pretty obstructive and noisy to me, but it's not awful.

brson:
Another concern. The install page still mentions the `%USERPROFILE%\.cargo\bin` path on Windows, and I'm not sure it will be obvious what `%USERPROFILE%` is, nor is this syntax valid PowerShell, and istm we want to be encouraging PowerShell over cmd.exe.

jonathandturner:
A few thoughts:

* Seems pretty straightforward, so people aren't going to get lost (yay)
* I like that we're showing the date now for the release
* I suspect we're going to get pushback on not linking to the downloads for people who want them, even though they're just on a different page now.  Because reasons.
* My nit would probably be to have "Install" at the top be "Download" just for better at-a-glance parsing

retep998:
In powershell, the % syntax does not work. Instead you have to do either `&"$env:USERPROFILE\.cargo\bin\foo.exe"` or `~\.cargo\bin\foo.exe`. In cmd though you have to use % syntax.

retep998:
Another thing to point out, the msvc version of Rust currently is built against VS 2013. So if the user gets the VC++ 2015 build tools, that will only provide the 2015 redistributables, but not the 2013 redistributables so they'll need to be installed separately. A few people have run into this issue, so it's not just hypothetical.

https://www.microsoft.com/en-us/download/details.aspx?id=40784

aturon:
@brson The updates look great! My only nit is on the install page -- can we avoid the horizontal rule after the Install Rust header? Given that the rustup directions are in their own box, the extra rule is adding extra clutter, I think.

Otherwise, LGTM!

brson:
> My nit would probably be to have "Install" at the top be "Download" just for better at-a-glance parsing

I actually changed it in this pr from "Download" to "Install" thinking that "Install" was more correct. If e.g. this page was titled "Download Rust" it would seem kinda wrong to me. Why is "Download" easier to parse than "Install"?

> Another thing to point out, the msvc version of Rust currently is built against VS 2013. So if the user gets the VC++ 2015 build tools, that will provide the 2015 redistributables, but not the 2013 redistributables so they'll need to be installed separately for rustc/cargo to even run.

This is kind of troubling. It seems to me that it's not a lot worse than today where if they pick the (recommended) MSVC build they will run into the same problem. I don't think there's anything blocking us from static linking our msvc builds now so hopefully we can resolve that soon.

brson:
Rebased.

The only thing I changed this round was removing the the horizontal rule from the install page but @aturon I think it looks poor, the "Install Rust" looks out of place to me. I have not changed "Install" to "Download" because I'm not convinced of it yet.

jonathandturner:
By parse I mean visually parse.  In general, I suspect the word "Download" will catch your eye faster than "Install". YMMV.

brson:
The look without the hr is growing on me, but I still think it doesn't look great.

aturon:
@brson I think it's good enough to ship. We can keep playing with these fine details over time.

brson:
I think I'm going to change the "Install" page back to "Downloads" after all, the deciding factor being that the front page has both "Install" and "Install Rust 1.13.0" links in big letters. Having the menu say "Downloads" instead will make the button the obvious choice. I think the title of the actual page still has to be "Install Rust" since it makes more sense than "Download Rust" considering the page's content. So the "Download" menu item will go to the "install.html", "Install Rust" page. Weird.

brson:
Argh, I can't decide.

brson:
OK, I still have it as "Install" for now. I've updated all the translations to use the new content from the english pages.

cc @aka-demik @3442853561 @thebergamo @KineticCookie @Calinou @lazpeng past translators there's a lot of content changing here soon.

aturon:
Looks fantastic. Let's land this thing!

steveklabnik:
🎊

aturon:
\o/

PR #627, opened 2016-12-03
fix links on the localized page

3442853561:
fix links on the localized page
fix the logo on the localized page


brson:
Thanks @3442853561 !

PR #629, opened 2016-12-04
[fixed and add some new]Chinese translation

3442853561:
- Fixed some links
- Modified some of the translation is not rigorous
- Translation of some new things


brson:
Thank you again @3442853561 !

PR #637, opened 2016-12-17
Translate new install.html to pt-BR

leodasvacas:
Translation to pt-BR of the new `install.html`. I'm feeling lazy about translating `other-installers.md` so if anyone wants to pick that up feel free. Also snuck some minor adjustments. Pinging @lazpeng for review.

### DIFF
--- a/_layouts/pt-BR/basic.html
+++ b/_layouts/pt-BR/basic.html
@@ -7,7 +7,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{{ page.title }}</title>
     <meta name="keywords" content="Rust, Rust programming language, Linguagem de programação Rust, rustlang, rust-lang, Mozilla Rust">
-    <meta name="description" content="Uma linguagem de programação de sistemas que roda incrivelmente rápido, previne segmentation fauls e garante segurança entre threads.">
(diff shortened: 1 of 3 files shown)


lazpeng:
Nice.
Should have a comma after "English" in "Nosso site em outros idiomas", located in _layouts/pt-BR/basic.html, if you could fix that and add it to the PR, I'd be very grateful. Also, "executar" may be better than "rodar", what do you think?

leodasvacas:
Made suggested changes. In some places "rodar" seems better in others "executar" seems better, I changed it in some places, what do you think now?

lazpeng:
Yeah, that's better, thanks.

aturon:
r? @steveklabnik 

In general, how would you like to handle ongoing reviews of translation PRs?

steveklabnik:
> In general, how would you like to handle ongoing reviews of translation PRs?

Usually, I try to check up on `rust-www` at least once a week to batch it up a bit. It's been a complicated last couple of weeks though. Often, I'll try to CC the people who made the original translation onto changes if they're more than something tiny, but at the moment, the process is effectively 100% trust based.

steveklabnik:
(in this case, @lazpeng is the person I'd want to ask here, so let's merge!)

PR #638, opened 2016-12-17
Add whitespace to fix markdown heading.

frewsxcv:
The heading "IRC Channels" in https://www.rust-lang.org/en-US/community.html is unformatted, displaying the raw `##` characters.


brson:
Thanks @frewsxcv !

Issue #640, opened 2016-12-18
Unrecognized platform


zeldin:
I got the following from the "Install" page:
```
I don't recognize your platform.

Rust runs on Windows, Linux, Mac OS X, FreeBSD and NetBSD. If you are on one of these platforms and are seeing this then please report an issue, along with the following values:
navigator.platform:
Linux ppc64
navigator.appVersion:
5.0 (X11)
```
So as requested I'm reporting an issue.  I'm on Gentoo Linux.

brson:
Thank you!

Issue #639, opened 2016-12-18
NetBSD is not recognized.


iquiw:
"I don't recognize your platform." message is displayed on [Installation page](https://www.rust-lang.org/en-US/install.html)

Here is the values the page shows.

```
navigator.platform:
NetBSD amd64
navigator.appVersion:
5.0 (X11)
```

I could install `rustup` and `rustc`, etc. by manually downloaded `rustup-init`.

brson:
Thank you!

PR #643, opened 2016-12-20
Add browser detection for Linux ppc64 and NetBSD

brson:
r? @steveklabnik

Fixes https://github.com/rust-lang/rust-www/issues/640
Fixes https://github.com/rust-lang/rust-www/issues/639


steveklabnik:
:shipit: :heart:

PR #651, opened 2016-12-22
Spanish translation

alx741:
So I started the Spanish translation here, but wanted to be sure I'm doing it correctly before proceeding with the rest of it.

I noticed the convention is `lang`-`2-letter-country-code`, but there is also french named just `fr` and the Spanish translation also fits best just as `es` (is that ok?)

### DIFF
--- a/es/friends.html
+++ b/es/friends.html
@@ -1,0 +1,71 @@
+---
+layout: es/basic
+title:  Amigos de Rust &middot; El lenguaje de programación Rust
+---
+
+<div id="users">
+
+  <h1>Amigos de Rust</h1>
+  <h2>(Organizaciones que ejecutan Rust en producción)</h2>
+
+  {% for user in site.data.users %}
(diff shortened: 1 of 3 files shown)


brson:
Thanks @alx741 ! It looks good to me.

Why is plain 'es' appropriate? I don't recall why 'fr' is used instead of something more specific. Maybe because there's overwhelmingly one type of French? @steveklabnik probably remembers.

alx741:
I now nothing about the best practices about this i18n naming, but I thought `es` fits better because I'm using Spanish that complies with [Real Academia Española](http://www.rae.es/) and no linguistic regionalisms whatsoever, so making `es-EC` (for *Ecuador*) would imply the possibility of adding `es-AR`, `es-ES`, `es-CO`, `es-MX`, and a dozen more with no differences between each other.

brson:
Thanks @alx741 !

PR #659, opened 2016-12-31
Fix French translation.

AndrewBrinker:
This commit basically reverts the locale name change done in #561. The pattern with other locales has been to use the longer-form IETF language tag for a given language (i.e. "en-US" vs. "en"). This leaves the door open for translations in non-majority dialects, and is generally more future-friendly. The reasoning given for moving from "fr-FR" to "fr" was that the change is consistent with Mozilla policy, but I think that the more important consistency is consistency with the rest of the Rust site.


AndrewBrinker:
For further background, [here](https://github.com/rust-lang/rust-www/pull/296) is the original discussion on using the full IETF language tags for locale names.

steveklabnik:
@AndrewBrinker 👍 , would you mind rebasing this?

AndrewBrinker:
Okay, I did not do that correctly. Just a minute.

Also, it looks like the Spanish translation now has the same problem as the French translation. If we're going to use the full language codes, we shouldn't accept PRs that don't.

steveklabnik:
> If we're going to use the full language codes, we shouldn't accept PRs that don't.

This is a good call; we shouldn't, and I should pay attention to that when reviewing. Thanks.

AndrewBrinker:
I'm going to close this. I've opened #682 which addresses the `fr` and `es` language codes.

AndrewBrinker:
Side note: thanks for taking translations so seriously! :smile:

PR #665, opened 2017-01-02
fix(font): use Mozilla version of font to fix bugs associated with Fira Font

### LINKED ISSUE #658: extremely light font rendering in chrome Version 55.0.2883.95 (64-bit)
screen shot 2016-12-29 at 4 28 49 pm

this may be due to a recent chrome update. ran into a similar issue with the npm website. can try to make a PR later today.

ashleygwilliams:
fixes #658 

pointing styles at the mozilla cdn for Fira Sans font fixes all display issues. replacing the woff files that were included with the ones from the mozilla cdn didn't. so i assume that there are different formats being used to fix the bug here. 

could probably pull everything in locally if we wanted, but it'd be bigger than what ya'll had before. lemme know if you'd like it local.


steveklabnik:
Removing the local versions seems fine for the website, but the equal fix in rustdoc would want the local files, I'd imagine.

ashleygwilliams:
ok i just added the local files. all come from here, fyi https://github.com/mozilla/Fira/blob/master/fira.css

brson:
Er

PR #669, opened 2017-01-03
new style languages links

3442853561:
There are more and more localized pages. The new style of links can make the page clear.


AndrewBrinker:
Is it necessary to make the language selector rely on JavaScript? Is there something about the text links that wasn't working? Text links have the advantage of being dead-simple, and working nicely with assisting technology like screen readers.

3442853561:
@AndrewBrinker these links make the page messy. There will be Japanese, Turkish,  
Spanish, Vietnam pages in the future.

AndrewBrinker:
Sure, there will be more choices, but I still think the current solution is the better one. It may be that there is a better way to design this, but requiring JavaScript for this basic functionality feels like a step back.

3442853561:
@AndrewBrinker What about creating a page for a localized pages links?

PR #671, opened 2017-01-04
Update french translation of install.html

lise-henry:
I noticed that the new install page wasn't translated in french. This is an attempt to work at this. Hope this is the right way to do it!


nikomatsakis:
Looks good! Thanks =)

PR #676, opened 2017-01-07
Update rustup platform detection

xen0n:
This is rust-lang-nursery/rustup.rs#866. Please don't merge until the MIPS rustup binaries are confirmed published.


brson:
Woo! Thanks @xen0n !

PR #682, opened 2017-01-11
Fixed language codes for fr-FR and es-ES

AndrewBrinker:
Instead of #659.

This PR renames the `fr/` language code to `fr-FR/` and does the same for `es/` to `es-ES/`. It also adds missing layouts for `es-ES/`, although they are currently just copies of the default layouts.

PR #684, opened 2017-01-11
Fix Spanish translation

alx741:
Wow! I failed to create a `_layouts/es/` for it and breaked the page, sorry about that! This will fix it.

### DIFF
--- a/es-ES/contribute.md
+++ b/es-ES/contribute.md
@@ -1,50 +1,50 @@
 ---
 layout: es-ES/default
-title: Contributing to Rust &middot; The Rust Programming Language
+title: Contribuyendo a Rust &middot; El Lenguaje de Programación Rust
 ---
 
-# Contributing to Rust
+# Contribuyendo a Rust
 
-You've started learning Rust. You love it, and you want to be a part
-of it. If you're not sure how to get involved, then this page
-will help.
+Has empezado a aprender Rust. Te encanta, y te gustaría ser parte del
+movimiento. Si no estás seguro como involucrarte, entonces esta pagina te
+ayudará.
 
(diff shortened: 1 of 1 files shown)


AndrewBrinker:
This is also addressed in #682, which also renames es/ to es-ES/, for consistency with the two-part language codes used elsewhere on the site.

alx741:
mmhh... In that case that should be merged first, then this PR could be rebased from there and merged as well so the `_layouts/es` has the Spanish translation instead of just the default English one.

brson:
Looks like the other is merged and this is ready for a rebase.

alx741:
Perfect, that should do it

alexcrichton:
@alx741 it looks like the files are now located as `es-ES`, want to update that location instead?

(sorry for the delay!)

alx741:
Sure, Now it should be ok

_layouts/es-ES/basic.html
```diff
@@ -21,24 +21,31 @@
...
-        {% include footer.html %}
+      <p>Nuestro sitio en otros idiomas&nbsp;:
+        <a href="/en-US/">English</a>,
+        <a href="/fr/">Francés</a>,
+		<a href="/ja-JP/">日本語</a>,
+        <a href="/ko-KR/">한국어</a>,
+        <a href="/pt-BR/">Português</a>,
+        <a href="/ru-RU/">Русский</a>,
+        <a href="/vi-VN/">Tiếng việt</a>,
+        <a href="/zh-CN/">简体中文</a>
```
alexcrichton: Oh I think this is now generally in `footer.html`, so perhaps just the "Nuestro ..." phrase could be updated and otherwise `include footer.html` could stay?

PR #685, opened 2017-01-11
Thread safety -> Data race safety

Manishearth:
Someone pointed out that this might be misleading, and they're correct.

r? @steveklabnik

### DIFF
--- a/fr/index.html
+++ b/fr/index.html
@@ -6,9 +6,9 @@ title: Langage de programmation Rust
     <div class="row pitch-row">
       <div class="col-md-8">
         <p class="pitch">
-          <b>Rust</b> est un langage de programmation système ultra-rapide,
-          qui prévient les erreurs de segmentation
-          et garantit la sûreté entre threads.
+          Rust est un langage de programmation système ultra-rapide, 
+          qui empêche les erreurs de segmentation
+          et garantit la sûreté des accès concurrents. 
           <br/>
           <a href="https://doc.rust-lang.org/book/README.html">Montre-moi&nbsp;!</a>
         </p>
--- a/en-US/index.html
+++ b/en-US/index.html
@@ -9,7 +9,7 @@ title: The Rust Programming Language
           <b>Rust</b> is a systems programming language
           that runs blazingly fast,
           prevents segfaults,
-          and guarantees thread safety.
+          and guarantees freedom from data races.
           <br/>
           <b><a href="friends.html">See who's using Rust.</a></b>
         </p>


steveklabnik:
So, we had talked about this, but nobody knows what "data race safety" means.

/cc @rust-lang/core

Manishearth:
"freedom from data races"?

steveklabnik:
"data race" itself is such a technical term that even people that use it often slip up and say "race condition"

Manishearth:
Hell, Wikipedia redirects data race to race condition.

fr/index.html
```diff
@@ -8,7 +8,7 @@
         <p class="pitch">
           <b>Rust</b> est un langage de programmation système ultra-rapide,
           qui prévient les erreurs de segmentation
-          et garantit la sûreté entre threads.
+          et garantit la liberté de accès concurrent .
```
Manishearth: not entirely sure if this is race condition or data race. Seems to be the latter (and is used below to mean that)

p0nce: `et garantit l'absence d'accès concurrent` would be appropriate.

Manishearth:
Another point brought up was that "prevents segfaults" is a negative statement and also a bit specific, instead using "provides memory safety" might be better and more positive. I'm not too sure of this, but it's worth discussing.

fr/index.html
Manishearth: Good catch! Totally forgot that `absence` was a word.

p0nce: While I'm here and bitching, the french frontpage says:
```
Rust est un langage de programmation système ultra-rapide, qui prévient les erreurs de segmentation et garantit la sûreté entre threads. 
```
This feels very odd, the correct sentence would be 

```
Rust est un langage de programmation système ultra-rapide, qui empêche les erreurs de segmentation et garantit la sûreté entre les threads. 
```
even better :

```
Rust est un langage de programmation système ultra-rapide, qui empêche les erreurs de segmentation et garantit la sûreté des accès concurrents. 
```
(guarantees safety of concurrent access)

p0nce: "Prévenir" really means "to notify"

p0nce:
As an input, I consider the punchline of a language to be about vision (what you value), rather than "Language does specific things X, Y and Z". Though the current one conveys the vision well

fr/index.html
Manishearth: Ah, good point. Would "et guarantit la sûreté des accès concurrents" be better than "et guarantit l'absence d'accès concurrent"? I'm not familiar with french programming terminology so I don't know if "data race safety" makes sense in French or if "freedom from data races" is better.

p0nce: Indeed "et guarantit la sûreté des accès concurrents" is bettter than "et guarantit l'absence d'accès concurrent" 
"accès concurrent" alone is ambiguous, it doesn't say if it's a successful access or a data race, might be both.

I'd say the most natural is
"data race safety" <> "sûreté des accès concurrents"

"data race" alone has no good translation, it would be translated by "race condition" but then the translation "condition raciale" which is official is a bit ridiculous-sounding and you should not use it.
So, er, both "data race" and "race condition" have not good, unambiguous, french translation.

alexcrichton:
This has been previously proposed and discussed: https://github.com/rust-lang/rust-www/pull/213

Notably @aturon [commented](https://github.com/rust-lang/rust-www/pull/213#issuecomment-154128559):

> Originally we had something about data races, which is what we actually provide a formal guarantee about (but is rather technical and not as compelling of a pitch).
>
> I don't think the change here does much to clarify things, unfortunately.

as well as @brson [commenting](https://github.com/rust-lang/rust-www/pull/213#issuecomment-154187252):

> I don't remember all permutations of this phrase but I think we settled on 'thread safety' as a succinct lay-explanation; people know or can intuit 'thread safety', not 'data races'. Also, as Niko points out in the thread, there is no commonly-accepted definition of thread safety.

As this has been discussed before and declined, I'm going to close this as well.

PR #694, opened 2017-01-20
chore: remove conflict tags

zhuangya:


alexcrichton:
It looks like this may no longer have any changes in it, I wonder if this is already fixed on master?

zhuangya:
yup closing it.

PR #701, opened 2017-01-25
Add an embedded chat

skade:
This adds instant access to #rust and #rust-beginners through our
website.

### DIFF
--- a/chat.md
+++ b/chat.md
@@ -1,0 +1,3 @@
+---
+layout: redirect
+---
--- a/en-US/chat.md
+++ b/en-US/chat.md
@@ -1,0 +1,16 @@
+---
+layout: default
+title: Rust Language Chat Rooms
+---
+
+<iframe src="https://kiwiirc.com/client/irc.mozilla.org:+6667/#rust,#rust-beginners" style="border:0; width:100%; height:80vh"></iframe>
+
+<div class="container">
+
+  <section>
+    <h2>Code of Conduct</h2>
+
+    <p>The <a href="https://www.rust-lang.org/en-US/conduct.html">Rust Code of Conduct</a> applies to these channels.</p>
+  </section>
+</div>
+


aturon:
cc https://github.com/rust-lang/rfcs/pull/1865 -- I think something along these lines could be plausible if we decide to stick with IRC as our primary entry point, but it's a bit premature to take this step until we've settled the question on the RFC, I think.

skade:
Hm, I was under the impression that the RFC is not about discussing a primary access point, just adding additional venues.

The community teams discussion yesterday was that inaccessibility and mediocre usability of mibbit is a major source of confusion when participating in IRC, so this would be a usable fix that should be monitored. I'll take that to the RFC discussion and #rust-community.

Hoverbear:
This might be slightly off topic or out of scope but I was very impressed with https://kiwiirc.com/ as a web IRC client, and it can have similar embedded links.

skade:
With the RFC apparently on the way out, and IRC sticking with us for another cycle, shall we move forward with this (meaning that I'd actually make the embedding not look like banged in literal 3 minutes and provide proper linking from the rest of the page).

aturon:
cc @rust-lang/core

Given that we're likely to stick with IRC for the time being at least, I think it makes sense to do something here.

(When I tried this out myself, I found that it took a *very* long time to load up in Firefox. Are there multiple such embedded clients we should consider?)

skade:
I know of no other and on novemb.rs, I never had any issues.

Could you try here? http://novemb.rs/chat.html

alexcrichton:
This seems fine by me, although it'd be nice if the styling would match the rest of the site (but that's likely quite difficult)

skade:
@alexcrichton that would, to the state of my knowledge, mean to design and write a complete theme. I'm also not sure how usable that might end up :).

aturon:
@skade Ok, I tried it there and it worked quite well. It looks slick!

I'll add this as an agenda item to the core team meeting tomorrow and try to reach consensus or at least gather more feedback.

skade:
Okay, please give me a thumbs up/down afterwards, I'd like to polish this properly, then.

aturon:
@skade Sorry for the long delay -- we've had a very full core team agenda and it took until today's meeting to get back to this.

In any case, the consensus was that this change seems like an improvement (though there continue to be some technical problems we should sort through before landing), and making it easier to access IRC is a good goal for the moment.

That should *not* be taken to mean that Slack, or some other communication channel, is out of the question for the future.

skade:
> @skade Sorry for the long delay -- we've had a very full core team agenda and it took until today's meeting to get back to this.
> 
> In any case, the consensus was that this change seems like an improvement (though there continue to be some technical problems we should sort through before landing), and making it easier to access IRC is a good goal for the moment.
> 
Cool, I'll clean the corners. This is a literal 5-minute proposal implementation :).

> That should not be taken to mean that Slack, or some other communication channel, is out of the question for the future.
> 
That reflects my stance, too.

skade:
It seems like there are a lot of kiwiirc users on moznet currently:

```
Error Connecting (Closing link: (kiwitest@195.26.67.20) [Too many connections from your IP (global class). If you have a Mozilla LDAP account, you can get a temporary exemption here: https://iplimit.irc.mozilla.org/])
```

Anything we can do about that?

steveklabnik:
@skade  any updates here?

skade:
@steveklabnik I haven't been able to get in touch with anyone at mozilla about this. I'll try again, thanks :).

PR #707, opened 2017-01-27
POD -> trivially copyable

### LINKED ISSUE #706: FAQ: "plain old data" in C++ is not related to being trivially copyable
https://www.rust-lang.org/en-US/faq.html#does-rust-have-copy-constructors

Types which implement Copy will do a standard C-like “shallow copy” with no extra work (similar to “plain old data” in C++).

should be changed to

Types which implement Copy will do a standard C-like “shallow copy” with no extra work (similar to trivially copyable types in C++).

ubsan:
Note that I'm not changing other languages; someone else who knows them will have to do that.

Fixes #706

### DIFF
--- a/en-US/faq.md
+++ b/en-US/faq.md
@@ -1520,7 +1520,7 @@ impl Foo {
 Does Rust have copy constructors?
 </a></h3>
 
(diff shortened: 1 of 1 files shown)


aturon:
Thanks!

Issue #649, opened 2016-12-21
the date of the migration of beta to stable is missing


unclejack:
The website used to display the date when the beta would become stable. It's not shown after the switch to rustup based installation steps.

This was useful because you'd know when to look for a new release.

3442853561:
Don't worry, I've fixed it https://github.com/rust-lang/rust-www/pull/680

3442853561:
I've fixed it https://github.com/rust-lang/rust-www/pull/709

PR #680, opened 2017-01-10
Fixed issues #649

3442853561:
Fixed issues https://github.com/rust-lang/rust-www/issues/649
Add the date of the migration of beta to stable

PR #709, opened 2017-01-30
Fixed [issues #649] and [pull requests #680]

3442853561:
Fixed issues https://github.com/rust-lang/rust-www/issues/649
Fixed pull requests https://github.com/rust-lang/rust-www/pull/680
Add the date of the migration of beta to stable

### DIFF
--- a/en-US/index.html
+++ b/en-US/index.html
@@ -18,7 +18,11 @@ title: The Rust Programming Language
         <a class="release-button" href="install.html">
           <div class="release-version">Install Rust <span>{{ site.stable }}</span></div>
         </a>
-        <div class="release-date">{{ site.stable_date | date: "%B %-d, %Y" }}</div>
+        <div class="release-date">
+		    {{ site.stable_date | date: "%B %-d, %Y" }}<br/> 
+            The next version is {{ site.beta }} <br/> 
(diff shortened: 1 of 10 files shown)


alexcrichton:
cc @brson, you may want to take a look at this change

3442853561:
What is the opinion of you guys? Put "the date of the migration of beta to stable" on which page is better? Index or install? @alexcrichton  @brson

brson:
Hi @3442853561. Thank you for the pr and sorry for taking so long.

While I am ok adding this information to the website, there are things about how this information is displayed that I think can be improved.

Here's a screenshot on my machine: http://imgur.com/a/sok31

The layout here is unbalanced, with much information in a small space, and uneven whitespace. Furthermore, the 'next version' seems like relatively unimportant information, since experienced Rust users will just know the next version, and new Rust users don't need to know.

The 'next version' can also be incorrect in the cases of point releases. "Next scheduled release" might be a more accurate description.

I think this information is not vital enough to appear on the front page. A more fitting place would be the negative space on the install page, beneath version information displayed there.

steveklabnik:
It seems we don't have consensus to merge, and so I'm going to give this a close. Thanks @3442853561 !

Issue #692, opened 2017-01-19
Epiphany on FreeBSD: I don't recognize your platform


wezm:
On <https://www.rust-lang.org/en-US/install.html> it shows, "I don't recognize your platform", when visited in the default GNOME3 browser on FreeBSD ("Web" aka Epiphany 3.16.3).

navigator.platform:
""

navigator.appVersion:
"5.0 (X11; FreeBSD amd64) AppleWebKit/601.1 (KHTML, like Gecko) Version/8.0 Safari/601.1 Epiphany/3.16.3"

wezm:
Additionally when this happens there's no obvious way to get a the default installation instructions (`curl https://sh.rustup.rs -sSf | sh`, which work in this case).

PR #711, opened 2017-02-01
Improve browser detection and install instructions

brson:
Fixes https://github.com/rust-lang/rust-www/issues/692

Corresponding rustup PR: https://github.com/rust-lang-nursery/rustup.rs/pull/936

cc @wezm

r? @steveklabnik


brson:
The "unknown" install box is now stuffed full of text, with the default install instructions. It's not aesthetic, but this is a rare case.

brson:
This also fixes an error in the markup where a div was improperly nested in a p.

steveklabnik:
Looks good to me. @wezm what do you think?

wezm:
I don't have access to my FreeBSD machine right now to test but the code changes look good to me.

PR #716, opened 2017-02-04
Add a comma

pithonsmear:
It was missing :)

Not much of a contribution (sorry), but I think it should be there.


3442853561:
I think this is not just an issue within the English page
Localized pages are growing. So that they are difficult to maintain.
Here's a solution https://github.com/rust-lang/rust-www/pull/708

alexcrichton:
Fixed in https://github.com/rust-lang/rust-www/pull/708, but thanks for the PR!

PR #719, opened 2017-02-07
Fix alphabet order and and link miss

hngiang:
Fix alphabet order layouts and add link miss


3442853561:
Localized pages are growing. So that they are difficult to maintain.
Here's a solution #708

alexcrichton:
Fixed in https://github.com/rust-lang/rust-www/pull/708, but thanks for the PR!

PR #727, opened 2017-02-11
Fixed some different places between Chinese and English

3442853561:
Because the English page has been updated


alexcrichton:
Thanks!

PR #740, opened 2017-02-24
replace the example

3442853561:
Let the example more like a calculator
Better show the match

### DIFF
--- a/_includes/example.rs
+++ b/_includes/example.rs
@@ -1,22 +1,34 @@
 // This code is editable and runnable!
 fn main() {
     // A simple integer calculator:
-    // `+` or `-` means add or subtract by 1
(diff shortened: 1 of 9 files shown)


brson:
Hi thanks @3442853561!

I really want to make this example better. I am worried that this patch makes the example bigger vertically, and I am concerned about the impact on the layout of the front page. I'd prefer if the box stays the same size it is, and that the example fit entirely within it, without scrolling. I suspect this content is too long to accomplish that (but I have not checked).

Here's another approach: https://github.com/rust-lang/rust-www/pull/756

3442853561:
I think https://github.com/rust-lang/rust-www/pull/756 is better

PR #745, opened 2017-02-27
es_ES: Fixed broken links on header

espectalll:


steveklabnik:
Nice catch, thank you!

PR #746, opened 2017-03-05
Fix distribution page link in FAQ

iKevinY:
This link currently points to an empty binary file, but according to [this Reddit thread](https://www.reddit.com/r/rust/comments/4s050r/httpsstaticrustlangorgdist_disappeared/), it should actually point to `/dist/index.html`.


alexcrichton:
Thanks!

PR #750, opened 2017-03-10
Updated English user group list

ericfindlay:
Added new user groups from Meetup. Removed a couple of user groups which no longer existed. Checked the links. Updated the total number of user groups. I removed the total number of members because it would be too much work to update it, and anyway it changes too quickly.

If this PR goes through, I'll paste the same list into the other language pages and submit that too.

### DIFF
--- a/en-US/community.md
+++ b/en-US/community.md
@@ -149,9 +149,9 @@ held by members of the Rust community from various conferences.
 
 ## User Groups and Meetups
 
-There are more than 50 [Rust User Groups][user_group] worldwide in over 20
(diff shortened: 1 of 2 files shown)


steveklabnik:
Hey @ericfindlay ! This looks good, but I noticed your git info isn't synched up with your github account, was that intentional or accidental?

ericfindlay:
Hello @steveklabnik. Sorry, I don't understand what you mean. What git info?

steveklabnik:
It looks fine now. Maybe it was a github caching issue. Your avatar was the default github one instead of your own. Looks fixed now!

PR #752, opened 2017-03-11
Fix layouts and different parts of the English page 

3442853561:
Fix layouts and different parts of the English page


alexcrichton:
Thanks!

PR #753, opened 2017-03-11
Added new user groups to the lists in French, Italian and Japanese.

ericfindlay:
r @steveklabnik 

The Japanese list is correct - I've adjusted the numbers of groups and countries but removed details of total members, to align with English language version. I haven't changed the numbers for the Italian and French lists, and as I don't speak either language they really require checking by a native speaker.


steveklabnik:
Thanks!

Issue #754, opened 2017-03-11
More Welcoming Snippet In Rust's Home Page


tebanep:
This is issue is a suggestion for improving the first impression a person gets with Rustlang. I think the code example displayed in https://www.rust-lang.org should be simpler and smaller. After actually using Rust, I noticed that it was way easier than I expected coming from Python. It wasn't obvious at first sight though. So my recommendation is to put a more welcoming snippet in the home page, so that people get motivated to try out the language!

This is the first Rustlang code many people see (at least it was in my case): 

// This code is editable and runnable!
fn main() {
    // A simple integer calculator:
    // `+` or `-` means add or subtract by 1
    // `*` or `/` means multiply or divide by 2

    let program = "+ + * - /";
    let mut accumulator = 0;

    for token in program.chars() {
        match token {
            '+' => accumulator += 1,
            '-' => accumulator -= 1,
            '*' => accumulator *= 2,
            '/' => accumulator /= 2,
            _ => { /* ignore everything else */ }
        }
    }

    println!("The program \"{}\" calculates the value {}",
              program, accumulator);
}

Compare it to that of Go's: (https://golang.org)

// You can edit this code!
// Click here and start typing.
package main

import "fmt"

func main() {
	fmt.Println("Hello, 世界")
}

Or Python's: (https://www.python.org)

\# For loop on a list
>>> numbers = [2, 4, 6, 8]
>>> product = 1
>>> for number in numbers:
...    product = product * number
... 
>>> print('The product is:', product)
The product is: 384

Thanks for making Rust. It is an awesome language that I hope I will be able to partner with Python for CPU expensive tasks!

3442853561:
This is a wonderful and exciting idea.
This is some of my advice.
Print "hello" in each language.
In order to show Rust's friendly and inclusive.
```
fn main() {
    println!("Hello");
    println!("こんにちは");
    println!("您好");
}
```

tebanep:
Right! One thing I've noticed with Rust is that the website is translated to several languages (including spanish). Neither Python or Go has that. This makes Rust look much more inclusive and diverse regarding its community!

brson:
This is done. Thanks @tebanep !

tebanep:
Thanks @brson for taking our suggestion into account! Hope it helps in making Rust more appealing for newcomers!

Issue #180, opened 2015-08-28
RWIB Mission 6: A Better Example


brson:
[Link](https://internals.rust-lang.org/t/rusty-web-improvement-bureau/2572/1#example).

Come up with a new example that fits in the same space and exemplifies
the Zen of Rust. This is a newb's first impression of Rust and it
needs to be breathtaking.

Consider that we might have multiple examples serving different
purposes and put them on a carousel.

mdinger:
The best analysis of front page examples I've seen is listed [here](https://users.rust-lang.org/t/is-the-homepage-example-wrong-on-purpose/763/11). It's well worth looking at.

lambda:
As mdinger mentions, I wrote up [my thoughts on examples](https://users.rust-lang.org/t/is-the-homepage-example-wrong-on-purpose/763/10) earlier; and I had done that research with the intent of proposing some front page examples, which I then didn't get around to. That said, I am still interested in the project; I have several ideas for examples, and just need to flesh them out.

As I proposed earlier, I think it would be best to have several small to medium sized examples, each focusing on a particular feature, rather than trying to have one be-all and end-all example. I think it's too hard to have a 20 line example that shows off all of what Rust is, and anything much longer will be too much to take in at once. Rather, I liked [Scala's approach](http://www.scala-lang.org/) of having a few feature bullet points, and each one show an example, though I would prefer to have one on display on the first load rather than having to click through to see any approach.

One sticking point I ran into when trying to write up some examples is that many of them require cargo or other things you can't use on Playpen. I think that it might be OK to have only some of the examples be runnable, while the others are just plain text that you can run if you download and install Rust and Cargo. Those that can be run on Playpen should be, but I think it's OK to not constrain ourselves that way.

Here are the current bullet points of Rust features on the front page; I am thinking of trying to write one example of each:

* zero-cost abstractions
* move semantics
* guaranteed memory safety
* threads without data races
* trait-based generics
* pattern matching
* type inference
* minimal runtime
* efficient C bindings

I think it would also be good to list Cargo, and how easy it is to use a fairly extensive ecosystem growing in Cargo.

For "zero-cost abstractions", it would be nice to embed something like the [compiler explorer](https://rust.godbolt.org/) to show an example of using iterator combinators that compile down into the equivalent of a `for` loop.

[...]

PR #756, opened 2017-03-14
Change example.rs to a simpler and more welcoming one

tebanep:
This pull request changes the Rust's website example to a more welcoming and simpler one according to issue #754 and as suggested by @3442853561.

This will make Rust seem as an easier language and encourage more people to use it!


brson:
Thanks @tebanep. This is something I've hoped someone would fix for a long time.

I like the idea of using a multicultural hello world, but am concerned that this example doesn't show much of Rust's character.

I could possibly persuaded that even with that drawback this example is better than the current one.

@steveklabnik do you have opinions?

Here is another thread on the subject to look to for ideas: https://github.com/rust-lang/rust-www/issues/180

steveklabnik:
My opinion is the exact same as yours: I love everything about this _except_ that it doesn't really show off anything about Rust.

> I could possibly persuaded that even with that drawback this example is better than the current one.

Yes, I think we should try to improve this, but if we really can't, let's merge anyway.

@tebanep any idea for something that might be even slightly more complicated? A boring but slightly better example might be putting these strings in a vector and then looping to print? Hm.

tebanep:
Thank you guys for considering this suggestion! Following @steveklabnik recommendation the example now uses an array and a for loop. What I basically did was to write the example as I would in Python and let the compiler point me to the errors and how to fix them. It worked flawlessly! I'm liking Rust very much!

3442853561:
This is a good idea! BTW, I guess "Bonjour" means "hello" in French.

3442853561:
What about this example?
```
fn main() { 
    let greetings = ["Hello",
                     "Hola",
                     "こんにちは",
                     "您好"]; 
                     
    for (num,greeting) in greetings.iter().enumerate() { 
        println!("{}", greeting);
        match num {
            0 =>  println!("This code is editable and runnable!"),
            1 =>  println!(""),
            2 =>  println!("このコードは編集して実行出来ます！"),
            3 =>  println!("这个代码是可以编辑并且能够运行的！"),
            _ =>  {},
        }
    }
} 
```

brson:
I'm happy with this as-as, though I like @3442853561's too. I'm going to just merge this and see how it looks live to make project, but I'm happy to keep considering improvements.

brson:
Thanks again @tebanep !

PR #779, opened 2017-04-09
Refinement wording for Chinese example at index page

eyeplum:
This pull request applies a small refinement for the Chinese example at the index page.

Detailed explanation:

- `这个` in Chinese means `this` (or more often means `this object`).
- `这段` in Chinese means `this piece of ...`, and is commonly used when referencing source code.


alexcrichton:
Thanks!

PR #780, opened 2017-04-10
Update index.html

vasily-kirichenko:
Better wording. "безумно" can be translated as "crazy" and sounds non-professionally. "невероятно" can be translated as "incredibly", which sounds more formally and appropriate for a front page.


alexcrichton:
Thanks!

PR #781, opened 2017-04-11
Fix website formatting issue

HadrienG2:
The Markdown engine used for rendering the rust-lang website does not like headings which are not preceded by a linefeed, i.e. this code...

```
Something
## A heading
```

...will be badly rendered as `Something ## A heading`.

Note that this behaviour violates the CommonMark spec ( http://spec.commonmark.org/0.27/#atx-headings : "ATX headings need not be separated from surrounding content by blank lines, and they can interrupt paragraphs"). So if you know more than me about the rendering engine in use, you might want to ask them to fix their parser.

But since putting blank lines before and after headings makes the markdown coding more consistent/readable anyway, I've went for the alternate option of working around the parser bug and putting line feeds everywhere :)

### DIFF
--- a/fr-FR/user-groups.md
+++ b/fr-FR/user-groups.md
@@ -4,6 +4,7 @@ title: Groupe d'utilisateurs Rust &middot; Rust, le langage de programmation
 ---
 
 # Rust User Groups
+
 Dans le monde entier, il existe plus 50 groupes d'utilisateurs Rust représentant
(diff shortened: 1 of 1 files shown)


steveklabnik:
> So if you know more than me about the rendering engine in use, you might want to ask them to fix their parser.

It's the default github stuff. kramdown?

> But since putting blank lines before and after headings makes the markdown coding more consistent/readable anyway, I've went for the alternate option of working around the parser bug and putting line feeds everywhere :)

agree 100%. Thanks!

PR #782, opened 2017-04-12
Change the example on the localized page to the latest version

3442853561:
Change the example on the localized page to the latest version


steveklabnik:
Thanks!

PR #784, opened 2017-04-20
adding TreeScale logo as a Rust User

tigranbs:
Our distributed PubSub and storage system written entirely in Ruts + MIO combination, which gave us 4-8X performance over Docker Hub and Quay.io for storing and processing Container images.
Would be happy to share our use-case!


alexcrichton:
Holy cow that's awesome! Thanks for the PR!

If you do a write up please feel free to share with use, we'd be more than eager to learn about it :)

PR #786, opened 2017-04-25
Add Rust Bangalore as requested

badboy:
Organizers requested to be added.

I only added to the english page for now as I saw there are multiple differences across languages. It might be useful to keep the full list of user groups in a single location instead of having it duplicated (and concurrently mutated) in multiple locations.

I can submit a PR to merge it back into one document and include redirects again if wanted.


steveklabnik:
Thanks!

PR #830, opened 2017-05-30
Fixed some Chinese translations

3442853561:
Fixed some parts which is the native speaker are not used to
Fixed the wrong header (friends.html)


alexcrichton:
👍

PR #838, opened 2017-06-06
Fix Italian greeting on example.rs

simonewebdesign:
In Italian language, "ed" is used whenever a word begins with a vowel.

Also removed a trailing space.


alexcrichton:
Thanks!

PR #848, opened 2017-06-13
Add links to some projects in Rust to the home page

carols10cents:
I got a suggestion recently that I think is awesome: we should show people the kinds of things people are making in Rust.

I've added links to 4 open source projects here-- I'd like to propose we keep this list to open source projects so that people could go see the Rust code if they wanted.

This is "playing favorites" somewhat, but I think we should be showing off our community's success stories!

r? @rust-lang/core

### DIFF
--- a/en-US/index.html
+++ b/en-US/index.html
@@ -52,6 +52,18 @@ title: The Rust Programming Language
       </div>
     </div>
 
+    <div class="row">
+      <div class="col-md-12">
+        <h2>What can you build with Rust?</h2>
+        <ul class="laundry-list">
+          <li>A <a href="https://servo.org/">parallel web browser engine</a></li>
+          <li>A fast, cross-platform <a href="https://github.com/BurntSushi/ripgrep">command line tool</a></li>
+          <li>An <a href="https://www.habitat.sh/">application automation framework</a></li>
+          <li>An <a href="https://www.redox-os.org/">operating system</a></li>
+        </ul>
+      </div>
+    </div>
+
   <script type="text/javascript">
     {% include include.js %}
 


alexcrichton:
Mind throwing up a screenshot of what this looks like as well?

I'm personally not a huge fan of playing favorites, although I'd be open to discussion here!

carols10cents:
No problem!

screen shot 2017-06-13 at 8 03 32 pm

nikomatsakis:
Hmm, I like the way that looks! We should definitely think about our "policies" here though, since I imagine people may ask to be included, and we'll have to decide how to handle it. =)

brson:
If we were to do this, I'd want to iterate more on the layout. The big white spaces, particularly to the right, look pretty awkward to me.

Since I'm not happy with how the current example overflows its box, it might make sense to rethink the layout of this page generally while we're making a big change to it. The two lists sitting side-by side might look better.

Havvy:
I'm personally against this change for the same reason as the feedback given in the image in https://twitter.com/dburka/status/883174678340268032 - the more clutter, the less we get our main message across.

Perhaps you can add it to that "See who's using Rust" page (https://www.rust-lang.org/en-US/friends.html), though that would also split the focus of that page. But that might be alright?

ranjithrajv:
Rather we can create a dedicated page like friends.html, an 'awesome list' to list all these exciting work by the community.

https://github.com/rust-lang/rust-www/issues/939

I'm working on it, anyone interested to join as a collaborator?

marioidival:
@aturon this can be merged? I like the @ranjithrajv idea, we can use [rust-unofficial/awesome-rust](https://github.com/rust-unofficial/awesome-rust) page to this...

steveklabnik:
We are re-doing the website, and so aren't interested in making big changes like this right now. Closing!

PR #863, opened 2017-06-30
Structured usergroups

badboy:
I extracted all user groups across all translated sub-pages (there were some inconsistencies) and pulled them into a single data file.
For the US I kept the states as we have them now (and fixed one or two).
I then modified all existing translated `user-groups.md` files to use the new data file (some languages still just use the english text).

The design is exactly the same as before.


badboy:
Force pushed and added missing Xalapa usergroup.

badboy:
ping @steveklabnik

PR #873, opened 2017-07-19
Add: indonesian translation

imamdigmi:
I added Indonesian for rust-lang.org site, so we Indonesians can easily learn Rust. Thanks!


steveklabnik:
Thank you! Awesome!

imamdigmi:
Thanks for your approval. There are still some pages that have not been translated yet I will fix them as soon as possible.

PR #875, opened 2017-07-19
ko-KR: Caught up with recent changes.

lifthrasiir:
This should be minor updates but synchronizes ko-KR materials to the most recent en-US contents.


steveklabnik:
Thanks!

PR #882, opened 2017-07-25
Fix spanish string for "This code is editable and runnable" in example.rs

plafue:
Although often omitted in texts and emails, Spanish exclamations should open with an opening exclamation mark.

See, for instance, [here](https://en.wikibooks.org/wiki/Spanish/Lessons/Introducci%C3%B3n_a_la_gram%C3%A1tica#Questions_and_Exclamations), [here](https://www.spanishdict.com/guide/spanish-punctuation#spanclasssdmdsdmdtextaccent2puntosdeexclamacinspanusesandexamples), and [here](https://www.thoughtco.com/upside-down-punctuation-in-spanish-3080317).

### DIFF
--- a/_includes/example.rs
+++ b/_includes/example.rs
@@ -8,7 +8,7 @@ fn main() {
         print!("{} : ", greeting);
         match num {
             0 =>  println!("This code is editable and runnable!"),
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks!

PR #883, opened 2017-07-25
Remove some obsolete links

matklad:
* We don't have AUTHORS.txt anymore, and there are over 2000 contributors :heavy_heart_exclamation:

* Subteam reports repo seems to be dead


steveklabnik:
Nice catch, thanks!

PR #885, opened 2017-07-26
Fixed MIR answer

### LINKED ISSUE #884: FAQ, "compilation seems slow", MIR
Hi, new here. At <https://www.rust-lang.org/en-US/faq.html#why-is-rustc-slow>, it mentions a hope of a future MIR optimization, but searching around it appears to me that MIR has already been merged. Has this optimization been implemented?

AndrewBrinker:
Resolves #884. If there's detail here that I am missing, let me know and I'll change the answer accordingly.


steveklabnik:
Looks good to me, thanks!

PR #891, opened 2017-08-09
Fix url of winkstreaming

0xpr03:
URL data missed a http:// creating 'https://www.rust-lang.org/en-US/www.winkstreaming.com' as URL


alexcrichton:
👍

PR #894, opened 2017-08-12
Add the Durham NC Rust user group

jbradberry:
We're called Triangle Rustaceans, and we meet every month on the 4th Monday.


alexcrichton:
👍

PR #895, opened 2017-08-12
French typography

eliovir:
A space is needed before the exclamation mark in French (see https://en.wikipedia.org/wiki/Exclamation_mark#French).


alexcrichton:
👍

PR #916, opened 2017-09-03
Switch IRC links to kiwiirc from mibbit

matklad:
Hi!

This PR proposes to switch IRC links from mibbit to kiwiirc. This should probably be discussed by the community team, but I don't know how to tag them, but probably @erickt knows :) 

Context:

I am among those people who don't feel comfortable with IRC, and for me, one of the greatest entry barriers was the absence of a good default web client. I've tried different clients (irccloud, mibbit, matix) unsuccessfully before I hit the kiwiirc, which makes IRC bearable for me (though I still don't know a ton of stuff, like how do you get those `* X thinks Y` messages, and how private messages work, if irc is sortof serverless, I would appreciate a link to a well-written short guide!).

So am totally not sure that kiwiircs will provide the best unboarding experience for all folks new to IRC, but I am pretty much sure that it's better than mibbit :) 

cc @jonathandturner @erickt


en-US/community.md
```diff
@@ -59,6 +59,9 @@ If you can't read English, You can also follow our [Weibo][weibo] for Chinese.
 
 Rustaceans maintain a number of friendly, high-traffic [IRC] channels on Mozilla's IRC network, irc.mozilla.org.
 
+If you don't know what IRC is, don't worry, it's easy to use: click any link to a particular channel,
+pick a username (you don't need to register, and you don't need a password) and that's it!
```
matklad: I also added a short blurb about how to actually use IRC. If someone adds a two paragraph tutorial here, I would be very grateful, because I still very much need one!

matklad:
Hm, just noticed https://github.com/rust-lang/rust-www/pull/701 which I've actually ❤️ a while back, but it seems stuck :(

steveklabnik:
related; @skade 's https://github.com/rust-lang/rust-www/pull/701

skade:
The problem is that I have no confidence that irc.mozilla.org doesn't block us regularly, because kiwiirc just has one source IP.

matklad:
ping @aturon :) Could you ask someone at Mozilla to exclude kiwiirc from IP limits? 

Sorry for being pushy about this, but it really saddens me that onboarding experience for non-IRC users like myself is quite poor, and that we haven't been able to improve it despite much talk :)

b-ambrus:
I support this change.  I have also tried Mibbit, and the reason why I don't like it is that whenever it sees something like `std::fmt::Debug` on irc, it displays that with the :D part replaced by a yellow smiley.  This makes programming discussions through Mibbit annoying, especially for new users who don't know how to disable that behavior.

Havvy:
FWIW, the smileys can be disabled by using a theme.

PR #917, opened 2017-09-04
Add the Schauspiel Dortmund as a production user of Rust

badboy:
Took me long enough, but here it is. The Schauspiel Dortmund (Theater Dortmund) is using Rust for a recent opera play. I built the small piece of software powering communication between light and sound systems on-stage. My employer allowed me to put that piece of software under an open-source license.

cc @skade, @eimann


skade:
\o/

alexcrichton:
👍

PR #925, opened 2017-09-16
Friends of Rust: ImageOptim

pornel:
I've been using Rust in production for a while and it works very well.


alexcrichton:
Thanks!

PR #926, opened 2017-09-16
Optimized images

pornel:
Reduced size of user-logos from 3MB to less than 1MB.

All images still look the same and have the same resolution.


pornel:
ping

alexcrichton:
Oops, thanks and sorry for the delay!

PR #930, opened 2017-09-19
Polish some zh-CN lines, makes it more idiomatic.

lochuan:
As a Chinese native speaker, I've been confusing with several lines of zh-CN version of Rust FAQ. 

> 我们不会夸奖其他目标的表现力、极简主义或优雅。这些是可取的，但不是主要目标。

Like this line above. I do not think majority of Chinese can get the meaning of this line clearly.


alexcrichton:
Thanks!

PR #931, opened 2017-09-20
Clarify how to subscribe

jethrogb:
I couldn't figure out how to subscribe in the Google Groups UI, so I googled it. Hopefully with this PR others won't have to.


steveklabnik:
Thanks!

PR #932, opened 2017-09-20
Add tromey to the dev-tools peers and move fitzgen and matklad to ful…

nrc:
…l membership

Also sort alphabetically

r? @aturon (could you also update the team email lists please?)


nrc:
Thanks @aturon!

PR #935, opened 2017-09-26
Remove outdated benchmarks faq

mgsloan:
I noticed that there's an FAQ entry saying that integrated benchmarks are considered unstable.  The entry comes from a year ago.  Is this still true?

Also, it seems like a weird question for a FAQ.  So even if it is still unstable, doesn't seem worth mentioning here.   It does not seem typical for a language to have an integrated benchmarking system.

### DIFF
--- a/en-US/faq.md
+++ b/en-US/faq.md
@@ -1491,7 +1491,7 @@ their new owners).
 How can I interoperate with C++ from Rust, or with Rust from C++?
 </a></h3>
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Yes, this is still true, and it does get asked frequently.

mgsloan:
Cool, I suppose it's not a weird question if asked from the perspective of "hey this isn't stable yet??".  Closing, thanks.

PR #937, opened 2017-09-28
Add German translation

fknorr:
First of all, credit to @medium-endian for doing at least half of the work, even if it's lost in the rebase!

All pages have been translated and all changes that happened since our fork were applied.

A few remarks:

- We used the informal/friendly "du" form of address instead of the professional/distant "Sie", which we think fits the community better.
- For consistency with the locale ordering (`de_DE`), German now appears as the first language in the footer - this might not be desirable.
- No attempt was made at keeping the phrasing gender-neutral everywhere, as this is notoriously hard to get right without sounding weird. If somebody is willing to look into this, please do not hesitate to correct us.

### DIFF
--- a/_includes/example.rs
+++ b/_includes/example.rs
@@ -1,24 +1,24 @@
 fn main() {
-    let greetings = ["Hallo", "Hello", "Hola", "Bonjour",
+    let greetings = ["Hello", "Hola", "Bonjour",
                      "Ciao", "こんにちは", "안녕하세요",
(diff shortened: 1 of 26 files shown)


alexcrichton:
Awesome, thanks!

I'd be fine merging this, but maybe the main example could be appeneded to instead of prepended to?

fknorr:
Like this?

alexcrichton:
Looks great!

PR #944, opened 2017-10-06
Improve Portuguese (PT[BR] / PT[PT]) sentence in example

Enet4:
In the Portuguese sentence of example.rs, the use of the Portuguese word _"esse"_ to mean the English equivalent _"this"_ may be common in Brazil, but in Portuguese from Portugal, the same word is only used to mean _"that"_.

This PR replaces _"esse"_ with _"este"_, so that it becomes compatible with both European Portuguese and Brazilian Portuguese.


alexcrichton:
Thanks!

PR #956, opened 2017-10-20
Solve rendering problem of bullet points in it-IT

gabibbo97:
This was bugging me out every time!
Guys, also, I would like to know if there is still some english that is not yet translated to italian in the repo.


alexcrichton:
Thanks!

steveklabnik:
>  Guys, also, I would like to know if there is still some english that is not yet translated to italian in the repo.

I'm not totally sure, but basically, check out the `it-IT` folder, and if there's no english there, then there's nothing left to do 👍

PR #970, opened 2017-11-01
Revert unintended change to TrafficLand logo on Friends page

bstrie:
In f77314a, "Optimize images", which changed many of the images on the
Friends page, the SVG markup was altered to give the TrafficLand logo a
blue background, which could not have been intended. Because the logo
was the same size before and after this commit (13 kb), and I cannot
otherwise imagine what was supposed to be optimized, simply revert the
change for that one file alone.


steveklabnik:
Thanks!

PR #991, opened 2017-12-22
Correcting fragment identifiers pointing to moderation team

davidalber:
The anchor for the Moderation Team on the [team page](https://www.rust-lang.org/team.html) is #Moderation-team (e.g., https://www.rust-lang.org/en-US/team.html#Moderation-team).

This PR fixes several instances where this is incorrect in the markdown. Most of those instances are #Moderation --> #Moderation-team (I guess the anchor was probably #Moderation sometime in the past).


alexcrichton:
Thanks! I think I may have accidentally created a merge conflict on one of the chinese translated pages though?

davidalber:
I rebased the branch onto master and resolved the conflict. Thanks!

PR #1018, opened 2018-01-24
New Website Logo: Zeplin

berkcebi:
We've just released our Adobe XD CC integration, which is built on Rust! 🦄⚙️

Organization contact: Berk Çebi, berk at zeplin dot io


berkcebi:
Thanks @ashleygwilliams!

Not sure how the process is on your end but let me know if there's something we need to add.

steveklabnik:
Just got a little lost in the shuffle, sorry about that!

Issue #924, opened 2017-09-14
Review the list of companies on the friends page to ensure each one isn't defunct, in preparation for the 100 friends milestone


bstrie:
We're nearing 100 companies on the friends page, and when we get there it would be nice to advertise that fact via some official avenues. But it will make us look silly if any of those companies are obviously defunct, and given that many of them are startups, the probability of a handful of them having gone under is high.

(For that matter it will make us look silly if any of them are no longer using Rust, but that's harder to determine at a glance. Making sure that a company still appears to exist via looking at its website is much easier.)

simeg:
This could even be automated :) The easiest approach would of course just be to curl all URLs and check the HTTP status code, and notify on != 200.

However if the company shuts down and the website is re-directed somewhere bad (think malware) then 200 would still be returned (like in #976) . Screenshots or selenium could be used. There's most likely software out there that does exactly this.

As a start a script could be written with the first approach and be part of the CI process.

PR #1020, opened 2018-01-24
Make sure websites of friends are up and running

simeg:
This idea sprung out of #924. I agree with @bstrie that it makes sense to have some sort of verification to make sure that the websites are up and running. The script cURLs all the URLs and if the response status code is 4** it prints a "WARNING", 5** it aborts and otherwise it just prints an "OK!".

```sh
$ time sh verify-websites-of-friends.sh
...
sh verify-websites-of-friends.sh  5.42s user 1.02s system 3% cpu 3:05.82 total
```

It will add ~3m to the CI process.

Let me know what you think.


bstrie:
I do like the idea of using a script for this, though I'm a bit wary at imposing 3 minutes of overhead just for checking the friends page. How much time does CI currently take? Alternatively we could just keep the script somewhere and run it offline once in a while.

simeg:
What's nice about the CI process is that it's automatic, so no need for anyone to do anything. Just wait an additional (worst case) 3 minutes. For what it gives you I think the overhead is worth it. I mean, how can we make sure that it gets run "once in a while"? Put our trusts in humans? 😄

simeg:
Pinging people who I've seen are active and might have valuable input to this.

(@steveklabnik @aturon @alexcrichton)

aturon:
This sounds great, thanks so much @simeg!

@alexcrichton look OK to you?

alexcrichton:
Could this perhaps be added in a separate matrix entry on Travis marked with being allowed to fail? I'd hate to block our own release b/c a different site happend to accidentally be down...

ashleygwilliams:
hey all, my two cents here:

i think that 3 min per build and a build failure signal are pretty high cost for the bit of information that a friend's site is down, potentially by accident (does not mean that they are actually out of business etc, would require another human check). i dont think this cost benefit tradeoff is worth it.

i do agree that having a computer do the work of checking if the sites are up is useful tho, just not as part of a push to production workflow. imagine this scenario:

- i write a patch that is super cool
- i make a pr
- pr fails because a friends site is down

does this block me? do we save this information somewhere?

chances are, we get the signal and we don't even fix it. i dont think fixing that entry belongs in my PR, and i would be loathe to make a contributor do that work. it's not fair and it doesn't really make sense.

ok so- what to do with this workflow?

i would recommend that perhaps we have this as either 1) a standalone test (yes it depends on humans! but... it always is gonna depend on humans) or 2) set this as a test that only runs on a branch (`deploy-friends`, `test-friends`) where you can push master to that branch and watch travis to see if it needs to be fixed. all in all, i want it to be something that a human opts into doing.

longterm, since there will be a LOT more friends, which means 3min turns into... 6min? 9min? this is not a future friendly solution as is. the test suite, manually opted in to, provides lots of time saving value without adding a huge overhead for little relative gain.

nikomatsakis:
might be nice to a general status page that we can check periodically?

ashleygwilliams:
hehe we've *almost* invented [nagios](https://www.nagios.org/) from first principles. perhaps i can talk with @aidanhs and the infra/ops team about something like this moving forward, but i would certainly say its out of scope of this PR ;)

simeg:
@ashleygwilliams You make a lot of good points. In my head the person finding the issue would fix it, but this being a repo with a lot of different contributors I certainly see the problems with that approach. I would not want people being blocked because of a website being temporarily down.

I like the idea about separate branch and the friends-badge. I'd be up for doing that work, but I'm not sure what the next step would be.

davidalber:
A third option, I believe, is to put the test into another repository and have Travis CI run it every twenty-four hours. That will avoid the special branch and triggering complications. You could still generate a badge and display it in this repository.

scripts/verify-websites-of-friends.sh
```diff
@@ -0,0 +1,56 @@
...
+
+    if [ "${status_code:0:1}" == "4" ]; then
+        echo "WARNING: URL [" "$url" "] responded with status code [" "$status_code" "], continuing";
+    elif [ "${status_code:0:1}" == "5" ]; then
+        echo "ERROR: URL [" "$url" "] responded with status code [" "$status_code" "], aborting";
+        exit 1;
+    else
+        echo OK!;
+    fi
+done
```
davidalber: Since this is I/O bound, you might try using [GNU parallel](https://www.gnu.org/software/parallel/), if you haven't already.

simeg: I did not consider this, will look into it!

simeg:
I think a separate repo would be nice, that way they wouldn't be coupled and you wouldn't have to worry about merging branches together. If you agree just create a new repo and I'll submit the PR there. Maybe there could be a global rust status repo? Not sure how you people like to structure things.

Also with the change that @davidalber proposed I managed to get the time down to ~1m 15s 👍

davidalber:
> Also with the change that @davidalber proposed I managed to get the time down to ~1m 15s 👍

Nice! Did you try it passing different values with `-j` (e.g., `-j 10`)?

simeg:
@davidalber Yup! Right now I'm working on making the code understandable.

Hopefully we can agree on where to put this. As I said before, a separate repo would be nice but I'm not familiar with how you structure things.

davidalber:
I'm not part of the crowd that can answer that question. If you are blocked by not having somewhere to set it up, you could create your own repository for starters. That would allow you to get everything set (including the Travis config, the build itself, a build status badge thing, a good README, etc.). I can't think of any reason (though, I could be overlooking something) that it wouldn't be simple to move it to a new home later.

simeg:
Finally got around to fixing this. I set up a repo with a Travis cron job that runs once every 24h and shows the status in the badge (_build passing/build failing_). Ideally this would not live on my repo but on a repo/branch under the Rust organisation, but a quick fix for now would be to display the badge in the README.

[Find the repo here](https://github.com/simeg/rust-www-friends-verifier)

Thoughts?

davidalber:
Nice! I think it makes sense to add a badge to this repository's README.md for increased visibility. The "travis example" badge in [rust-lang-nursery/rustfmt](https://github.com/rust-lang-nursery/rustfmt) is like that (i.e., it's generated by a build in a different repository). How do those of you with the ability to merge PRs here feel?

simeg:
Closing as no response in a long time.

PR #1026, opened 2018-01-30
Fix ambiguity between brackets and braces

faheel:
The section:
> I can leave out parentheses on <code>if</code> conditions, so why do I have to put brackets around single line blocks? Why is the C style not allowed?

mentions "brackets" instead of "braces" (or "curly braces"), which seems ambiguous as it doesn't specify which types of brackets are being talked about (curly, square, round or angle brackets).

I've changed "brackets" to "braces" in this section.

### DIFF
--- a/en-US/faq.md
+++ b/en-US/faq.md
@@ -250,19 +250,19 @@ Not in the typical sense used by languages such as Java, but parts of the Rust s
 
 <h2 id="syntax">Syntax</h2>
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Hi @faheel ! Sorry to take so long to review this!

We use "curly brackets" as the proper way to disambiguate here, would you mind changing to that please? Thank you!

faheel:
@steveklabnik I've made the changes.

PR #1027, opened 2018-01-31
Remove defunct Rust NY meetup group.

frewsxcv:


steveklabnik:
yay for the new one, at least!

PR #1033, opened 2018-02-05
Move Matklad to alumni status

nrc:
r? @aturon Could you remove Matklad from the dev-tools and cargo mailing lists too please?


aturon:
Done on both counts. 

@nrc, can you make a PR against https://github.com/anp/rfcbot-rs/blob/master/teams.toml to also update rfcbot?

PR #1035, opened 2018-02-05
Publish Tilde whitepaper! 🎉

carols10cents:
This publishes the first of the whitepapers I'm working on, about Tilde. /cc @wycats, who has previously approved this text, we're making it public!

Here are a few screenshots of how the front page looks with this new link at a few different widths, this isn't ideal but we have @ashleygwilliams working on a new website that will have a better place for whitepapers in general:

screen shot 2018-02-05 at 1 37 38 pm

screen shot 2018-02-05 at 1 38 33 pm

screen shot 2018-02-05 at 1 39 54 pm

The whitepaper isn't translated (yet?) so I didn't add links to the other languages' homepages... 

/cc @rust-lang/core and @booyaa for the content team... should we do fcp ticky boxes? Anyone have any objections to this going out?


ashleygwilliams:
this is awesome @carols10cents - one thought. to make this more future friendly would you mind:

- instead of linking directly to a PDF, link to a page called rust whitepapers or something
- link to the pdf on that page
- add a call to people who might want to do a whitepaper with us

this way we get some marketing to funnel in new white paper candidates, we have a place for more than one white paper, and we dont link to a file from the landing page (some people don't like that much)

thoughts? (also i am willing to help with that page's design and content)

carols10cents:
Great suggestion! I've added a bare minimum page, wording/content suggestions MUCH appreciated!

I also made the "reach out" link mail to the community team, with the thinking that this would be the content subteam's domain? Does that sound right?

Here's a screenshot of the rendered page:

screen shot 2018-02-05 at 2 08 33 pm

ashleygwilliams:
oh and to answer your question @carols10cents i think this should be the domain of the commercial end users group but since that group does not exist yet and won't official exist for a solid week or more we can fix that later

simeg:
@carols10cents This is awesome! 🙌 🎉

PR #1043, opened 2018-02-10
fix arabic translation for سلام message

adals:
"سلام" word is an Arabic word but text below are not in Arabic language at all and Arabic speaker will not understand it completely so here is my correction .


khaledkbadr:
I think that was supposed to be in Persian, not Arabic. But I would love to see Arabic since it's far more popular.

steveklabnik:
Yes, it is supposed to be persian. Would you mind adding the arabic, rather than replacing the persian? Thanks!

adals:
i have updated arabic language

adals:
why does this PR take so long to be merged? I did what you asked from me

Issue #1030, opened 2018-02-02
Reviewers not being automatically assigned


davidalber:
[CONTRIBUTING.md](https://github.com/rust-lang/rust-www/blob/master/CONTRIBUTING.md) states:

> All pull requests are reviewed by another person. We have a bot, @rust-highfive, that will automatically assign a random person to review your request.

Automatic PR review assignment is not happening in this repository. If this is deliberate, we should update CONTRIBUTING.md. If it's not deliberate, we should decide whether to fix it or to accept it and update CONTRIBUTING.md.

davidalber:
This has been fixed (presumably by rust-lang-nursery/highfive#86)!

PR #1047, opened 2018-02-13
Adding missing period

davidalber:
Several translations of the [install page](https://www.rust-lang.org/install.html) is missing a period at the end of one of its paragraphs. For example, from the [English translation](https://www.rust-lang.org/en-US/install.html):

> Rust is installed and managed by the rustup tool. Rust has a 6-week rapid release process and supports a great number of platforms , so there are many builds of Rust available at any time. rustup manages these builds in a consistent way on every platform that Rust supports, enabling installation of Rust from the beta and nightly release channels as well as support for additional cross-compilation targets

This PR adds the missing period.

### DIFF
--- a/en-US/install.html
+++ b/en-US/install.html
@@ -111,7 +111,7 @@ title: Installation &middot; The Rust Programming Language
           <code>rustup</code> manages these builds in a consistent way on every
           platform that Rust supports, enabling installation of Rust from the
           beta and nightly release channels as well as support for additional
-          cross-compilation targets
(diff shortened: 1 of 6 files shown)


davidalber:
Also, for full disclosure. I looked for something to fix so that I could open a PR and test if #1030 has been fixed. It has! 🎉

davidalber:
I asked highfive to reassign this since the "all" group of reviewers was recently changed.

PR #1051, opened 2018-02-15
Replaced the broken link to libs.rs with...

mark-summerfield:
one to crates.io (but it doesn't seem to work as a link, but at least it is the right thing to say).


ashleygwilliams:
hi @mark-summerfield ! in order to get the link to work as you have written it you'll need to add a link at the bottom of the page, something like `[crates.io]: https://crates.io/`. you'll probably also want to remove the link you are removing! lemme know if you need anything further help!

great find on the broken link btw! really appreciate this PR :D

mark-summerfield:
I've added the change you suggested but don't know what I'm supposed to do next.

ashleygwilliams:
great- this looks perfect except now we just need to remove the `[libs.rs]` link down on the bottom. once that is done i can merge this and it will be shipped! thanks again :D

PR #1072, opened 2018-03-21
Lang team updates

Centril:
Announced on internals: https://internals.rust-lang.org/t/new-members-for-the-lang-team/7087


aturon:
Hurrah!

PR #1075, opened 2018-03-26
update gems

steveklabnik:
Deployment failed the other day: https://travis-ci.org/rust-lang/rust-www/builds/356152471

i figured i'd do a gem update and see what happens...


Mark-Simulacrum:
I'm somewhat hesitant to land this before we have the release done -- just in case it breaks something.

IIRC, though, the failure you noted isn't actually persistent, it should be spurious.

PR #1081, opened 2018-04-03
Added install instructions for WSL users - fixed #889

### LINKED ISSUE #889: Install page is awkward for WSL users
When I go to https://www.rust-lang.org/en-US/install.html as a windows 10 user, I see an install link for Windows 10, which is great. 

However, Windows 10 now has the Windows subsystem for linux, which many devs use, and which is where I'd like to install Rust. From the Windows install page, the linux install instructions are (afaik) only accessible from the "other installation methods" section. 

Would it make sense to either a) add a more obvious link to installations for other platforms for all install pages or b) add a subsection for WSL install instructions for Windows users? 

a) might have the advantage that it would work nicely for anyone whose user-agent is improperly handled.

donaldpipowitch:
I just hit https://github.com/rust-lang/rust-www/issues/889 as well. I guess it becomes pretty normal for Windows developers to use WSL, so I think this change helps here to get started easily. ❤️

### DIFF
--- a/en-US/install.html
+++ b/en-US/install.html
@@ -17,6 +17,8 @@ title: Installation &middot; The Rust Programming Language
             <a href="https://win.rustup.rs">rustup&#x2011;init.exe</a>
             then follow the onscreen instructions.
           </p>
+          <p>If you're a WSL user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
+          <pre>curl https://sh.rustup.rs -sSf | sh</pre>
         </div>
 
         <div id="platform-instructions-unknown" class="instructions" style="display: none;">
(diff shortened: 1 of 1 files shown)


en-US/install.html
```diff
@@ -17,6 +17,8 @@ <h1 class="rustup">Install Rust</h1>
             <a href="https://win.rustup.rs">rustup&#x2011;init.exe</a>
             then follow the onscreen instructions.
           </p>
+          <p>If you're a WSL user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
```
ashleygwilliams: could you expand this acronym? otherwise this looks good to go! thank you!

ashleygwilliams:
thanks so much!

donaldpipowitch:
Thank _you_!

PR #1084, opened 2018-04-07
Added contribute community translation for zhCN

shingtaklam1324:
I added the translation for the rest of the page. I'd prefer this not to be merged until it has been checked by another Chinese Speaking user as my Chinese is not the best.


shingtaklam1324:
This is way out of data and probably pointless now. I don't use rust actively either so I can't really update it either.

PR #1096, opened 2018-04-24
Fix translation for zh-CN/other-installers.md

bombless:


bombless:
@ashleygwilliams You can see that there is an English sentence in the version I edited, so that sentence must just slipped to the translated version accidentally I believe.
Also for the translation for "Pass `--help` to `rustup-init` as follows to display the arguments	
-`rustup-init` accepts", the meaning of "as follows to display" was missing so the whole translation for this sentence was confusing.

bombless:
Closing since this repo has been deprecated

PR #1098, opened 2018-05-03
Team updates

ashleygwilliams:


aidanhs:
r+ on infra changes

oe:
hmm, is adding all subteam members to the community team in an umbrella fashion the way to go here? can't we do something akin to what the rustdoc team did?

ashleygwilliams:
hey @oe it's sorta complicated... one of the reasons i'd like to do it as this first is because i would like to work through https://github.com/rust-community/team/issues/232 before we do much more. the new website will have a much better structure- this is mostly a stopgap PR to make sure everyone is at least listed and then we'll either keep making adjustments here, or wait til the new site in august (which is not too far away!)

oe:
okay! just asking in case anyone overlooked the possibility, but it looks like you have good reasoning

_data/team.yml
```diff
@@ -305,7 +328,7 @@ teams:
   - name: Community team
```
oe: For the time being, this could say something like "Community team (including subteams)"

PR #1119, opened 2018-05-30
Add Threema to friends

dbrgn:
We've been using Rust on the server for a while, and today the first app beta containing Rust code [went live](https://threema.ch/en/blog/posts/threema-web-for-ios-beta) :slightly_smiling_face: 

As contact you can use my commit author e-mail. Is that sufficient, or should I also send the information to user-logos@rust-lang.org?


steveklabnik:
Congrats!

> As contact you can use my commit author e-mail. Is that sufficient, or should I also send the information to user-logos@rust-lang.org?

Nope, this is fine. Thanks!

PR #1140, opened 2018-06-29
Change "Fork me on Github" to "Fork us on Github"

LTheGreats:
The reason why I would make this change is because more than one person works on the rust project.


brson:
Interesting suggestion, but "me" in this case is anthropomorphizing the project itself. It's not referring to a person. "Fork me on GitHub" is a standard banner for GitHub projects.

LTheGreats:
okay

PR #1166, opened 2018-08-02
allow 10x Genomics logo to resize to fill div

pmarks:
drops the `width=48 height=48` attributes.


pmarks:
@steveklabnik ping -- who can approve this?

PR #1177, opened 2018-08-14
Remove myself from the infra team.

tomprince:
@aidanhs


aidanhs:
Thanks for the PR @tomprince! And I just want to say thanks again for the effort you put in, and good luck with your future work :)

PR #1188, opened 2018-08-29
added working group information

avadacatavra:
I know that the website is getting revamped, so if this is unnecessary please feel free to close it. I’ve gotten feedback that it’s difficult to see what WGs exist and what they’re working on. I figured even a short term solution might help fill this gap.


Deedasmi:
Off topic, but do you have a link for the revamp? Missed that news

PR #1201, opened 2018-09-26
Use T-rustdoc label instead of A-rustdoc label

nlopes:
The contribute to rust tools section uses A-rustdoc as the label to search for in github
but the label in github is T-rustdoc. This changes the documentation links to use
T-rustdoc per github label.


steveklabnik:
Thanks!

Issue #1190, opened 2018-09-07
Typos in example on the Rust home page


m-f-h:
Hi, although these are just minor typos, since this is prominently displayed on the main homepage (https://www.rust-lang.org/) and for each language, it might be worth fixing it:
- The Arabic phrase has the exclamation mark to the right, instead of the left.*
- The Czech phrase has no exclamation mark at all.

(*Additionally, maybe related, that "!" comes finally out on the very right of the output [supposed to be written in two steps from right to left, so that's slightly nontrivial...]. I assume somehow it is not in "RTL mode" in the string. But anyway, putting it to the left should (?) resolve the issue. 
[Hm. Having tried some edits, that seems to be a resistant problem... Whether I put a "!" to the left or to the right, in a separate string or in a format string with " {} " or " {} {} ", or whatever, it always end up to the right! This is weird!!]
So this example is a kind of disproof of the statement itself, sine the RTL text can actually (almost) NOT be edited, at least in my Chrome browser: juts try to delete that "!"...)

HTH.
-M.

zeb-stephen-downs:
I think this is an issue with the embedded editor they're using (third party called Ace (Ajax.org Cloud9 Editor)); if you notice, you can move the cursor 3 spaces to the left and hit delete and it will delete the ! mark. I tried something similar on the third party's own website and got the same results. I'm thinking this is an encoding issue with the editor.

m-f-h:
I agree, but this only concerns the side note (* ...), not the main bug report.

steveklabnik:
We are going to be re-doing the Rust website very soon, and it does not have this code example any more. I'm going to leave this open since it's technically a bug, and if anyone wants to fix it, please send in a PR, but it's likely to not get fixed until it's removed.

PR #1207, opened 2018-10-05
Update static code example

Bassetts:
Updates thes static code example on the index page to match the dynamic
 code example.


Bassetts:
As I was touching the example code I thought it made sense to add missing exclamation marks and to move the exclamation mark in the Persian example as per #1190.

PR #1223, opened 2018-10-29
Add Routific among friends

oddg:
Routific is using Rust in production. The experience is delightful and we would like to add our logo on your page.
You can contact me on gh or at `olivier@routific.com` if you have any question.


oddg:
Thank you!

PR #1245, opened 2018-11-29
Remove "click here" from flash banner and increase clickable area

wiktorkuchta:


ashleygwilliams:
thanks for this @wiktorkuchta ! could you include a screenshot? if everything looks good i'm happy to merge :D

wiktorkuchta:
image

The layout doesn't change on any screen size, only the text and clickable area. 
It's easy to see by using browser Dev Tools

